PR #13, opened 2020-06-22
Quadkey high zoom support

Algunenano:
Adds support to generate quadkeys for high zoom levels (should be ok up to z52).

Removes all the pixel logic since it's unnecessary .


jatorre:
Thanks @Algunenano

PR #17, opened 2020-09-17
link more s2 geometry js fonction to bigquery

francois-baptiste:


jatorre:
This is awesome!

One question, do you have available the coverage method? We got a request for that

PR #19, opened 2020-10-20
Ensure geographies don't go over the antimeridian

Algunenano:
Something like 311131313 got a x_max of 180.00000000000003, which is just sligthly
over the antimeridian, so when reading the GeoJSON BQ translated it to ~ -180 which
created a line across the world in the opposite direction of what we want


jatorre (approved):

jatorre:
Thanks.

PR #23, opened 2020-12-14
patch facePosLevelToId shouldn't return signed Long

SMazu:
Long.fromString(bin, true, 2).toSigned().toString(10) -> Long.fromString(bin, true, 2).toString(10)

// check: https://git.coolaj86.com/coolaj86/s2-geometry.js/src/branch/master/src/s2geometry.js#L480


SMazu:
image
```
'use strict';
 
var S2 = require('s2-geometry').S2;
 
var lat = 40.2574448;
var lng = -111.7089464;
var level = 15;

var key = S2.latLngToKey(lat, lng, level);
// '4/032212303102210'

var id = S2.latLngToId(lat, lng, level);
// '9749618446378729472'

var id = S2.keyToId(key);
// '9749618446378729472'
```

`key` is correct, `id` is wrong

francois-baptiste:
Yeah I agree with that fix, thank you @SMazu

francois-baptiste:
By the way we should also use uint64 for H3 cells representation, this format helps reducing data size!

SMazu:
awesome, can we merge this then?

jatorre:
Happy to merge! Thank you very much.

jatorre:
Oh going to need help

SMazu:
hi @jatorre @francois-baptiste, is there anything still pending for merging this?

jatorre:
Sorry is taking some time. We are working on a full rebase of jslibs to add continuos integration and other exciting things, so is taking a bit longer. But will try to push today.

jatorre:
Should be deployed too.

PR #24, opened 2020-12-17
Add ability to deploy to private GCP project

mwjackson:
Our GCP projects are behind VPC Service Perimeters, which unfortunately do not allow calling of the `jslibs` functions. This PR adds a secondary script `custom_deploy_functions.sh` that allows users to deploy a single copy of the library to a project & region of their choosing.

There are also a few minor amendments made to the naming/ordering of the `sql/` files that allow the script to run through without error.


jatorre:
Thanks for this, super useful.

@alasarr could you review this?. 
I was not aware users with VPC Service parameters can not run directly. Makes sense.

We should probably then start thinking on adding a version to the library for each deployment to check what version of jslibs you have installed

garrywilliams:
Thanks for providing this H3 functionality - I can confirm this update is required, and have written a similar script to deploy to our private GCP project and bucket (in the europe-west2 region)

jatorre:
Thanks for the contribution.

We are currently doing a entire refactor of the project that will bring documentation, CI and test. We will also include a system for deployment of the functions on custom installations. So this should take over this functionality.

jatorre:
Hey @Jesus89 can we revisit this? Right now I dont see documentation to imstall the spatial extension on your own BigQuery and this is a common use case.

Jesus89:
Hello. We have just merged a refactor of the entire project and the deployment to custom projects is already included.

1.  Copy and fill [this file](https://github.com/CartoDB/carto-spatial-extension/blob/master/common/bigquery/dev_env.sh.sample) to `dev_env.sh` with your custom data
2. Load those env variables: ```source common/bigquery/dev_env.sh```
3. Run `make deploy` globally or per module

PR #25, opened 2021-01-13
Amended script to build functions for Topo in eu_west1 region

garrywilliams:
Previous installation into topo-data team had used search-and-replace to update all of the SQL scripts. However, to minimise impact and the amount of change, the search/replaced happens on submission to BigQuery. This will make it easier to keep in line with changes made by the CartoDB developers and make the scripts more portable if we want to relocate them to a different project.


garrywilliams:
PR raised in error!

PR #31, opened 2021-02-12
Quadkey module

vdelacruzb:
Wanted to check with you before you go on holidays, at least the basics and name of functions.

- I created some integration tests using for loops but I set the step really high to avoid long waiting periods. The step in js checks is lower to compare more values.

- The doc is very basic. There is not much into these functions just some conversions and they are public functions so the doc is more for users and less for devs. Will have to iterate on it and see with product as they will generate also public documentation.

- I have to iterate in the implementation of parent and children, run out of time. They work fine, but it should be relatively easy to do it without decoding encoding.

- Also have to explore more deeply how the mapbox tile cover works. It converts GeoJSON to Quadkeys, and I guess it should be possible to directly convert to quadint and skip steps.


quadkey/README.md
```diff
@@ -0,0 +1,12 @@
+# Quadkey
+
+This contains all the necessary code to build, generate, test and deploy the [BigQuery User Defined Functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/user-defined-functions) for interacting with quadkey indexes.
```
Algunenano, line 3: This shouldn't mention BQ here, since we will have other database support.

quadkey/bq/Makefile
```diff
@@ -0,0 +1,77 @@
+# Programs
+SED ?= sed
+BQ ?= bq --location=$(BQ_REGION)
+GSUTIL ?= gsutil
+
+POST_INTEGRATION_CLEANUP ?= 1
+
+# Deployment variables
+BQ_LIBRARY_QUADKEY ?= $(BQ_BUCKET_PUBLIC)$(BQ_DATASET_QUADKEY)/quadkey_library.js
```
Algunenano, line 9: SKEL library is named `SKEL_BQ_LIBRARY` so this should be `QUADKEY_BQ_LIBRARY`

quadkey/bq/sql/BBOX.sql
```diff
@@ -0,0 +1,16 @@
+-----------------------------------------------------------------------
+--
+-- Copyright (C) 2020 - 2021 CARTO
```
Algunenano, line 3: Copyrights should be 2021 only, right?

Algunenano (commented):
Some notes from a quick preview (not in depth):
* There shouldn't be parseInt / Number calls in the JS functions just because BigQuery require them, that should be part of the BQ API, not the library itself.
* I see a LICENSE but i'm worried it's added ad-hoc and not the real license.
* Related to the previous point, where does the code from `quadkey.js` comes from. What part comes from other project and how do we update it? What part is added by us? How do we mix the 2 together?
* The style of the JS tests is completely different from the code. Maybe we should lint the tests too? And maybe a common .eslint file somewhere in the repo :thinking: ...

ibrahimmenem (approved):

vdelacruzb:
> * I see a LICENSE but i'm worried it's added ad-hoc and not the real license.
> * Related to the previous point, where does the code from `quadkey.js` comes from. What part comes from other project and how do we update it? What part is added by us? How do we mix the 2 together?

Regarding these two topics, I basically took the same code that we had in jslibs -> quadkey.js. And then adapted it to support quadints instead of quadkeys by updating the functions that used quadkeys and also by adding parser from quadkey to quadints. So it should be under the same terms of license than the past code. Was the last code taken from anywhere or totally created here? 

I would say that the quadint part doesn't have other sources. I calculated that we needed 5 bits for zoom and then when saw this https://github.com/joekarl/binary-quadkey it was a verification, but we didn't follow that github implementation.

Inside tilecover.js we use mapbox/tilecover. Does it affect our license?

quadkey/bq/test/quadkey_integration.js
```diff
@@ -0,0 +1,171 @@
...
+            const [job] = await client.createQueryJob({ query: query });
+            [rows] = await job.getQueryResults();
+        });
+        assert.equal(rows.length, 1);
+        assert.equal(rows[0].versioncol, 1);
+    });
+
+    it ('Should be able to encode/decode quadints at different zooms', async () => {
+        let tilesPerLevel, x, y;
+        for(let z = 1; z < 30; z = z + 7)
```
Algunenano, line 33: Please don't do a single query per zoom level, this is going to be awfully slow. Instead compose a query with all the things you want to check and send a single query to BigQuery and then check the results. You should also add tiles in between the first and last, and check what happens when invalid values are passed (if we want to handle it).

Same for ALL other integration tests.

quadkey/library/test/quadkey_test.js
```diff
@@ -0,0 +1,133 @@
...
+                    childs.forEach((element) => {
+                        if (currentChild === element) {
+                            ++cont;
+                        }
+                    });
+                    assert.equal(cont, 1);
+                }
+            }
+        }
+    });
```
Algunenano, line 132: Missing sibling tests.

quadkey/bq/test/quadkey_integration.js
```diff
@@ -0,0 +1,171 @@
...
+                    childs.forEach((element) => {
+                        if(currentChild === element)
+                        {
+                            ++cont;
+                        }
+                    });
+                    assert.equal(cont,1);
+                }
+            }
+        }
```
Algunenano, line 169: Where are the tests for all the other functions (`ST_ASQUADINTPOLYFILL`, `ST_GEOGFROMQUADINT_BOUNDARY`, `KRING`...)? Is there something missing in this PR?

quadkey/library/js/tilecover.js
```diff
@@ -0,0 +1,585 @@
+(function() {
```
Algunenano, line 1: Where is the copyright, license and notes about where this comes from? We don't have to add other stuff to files that come from other repositories if we copy them as-is, but we should be extremely clear about it in the license notices and document where they come from, what they do and how to update them.

Algunenano (changes requested):
> Regarding these two topics, I basically took the same code that we had in jslibs -> quadkey.js. And then adapted it to support quadints instead of quadkeys by updating the functions that used quadkeys and also by adding a parser from quadkey to quadints and viceversa. So it should be under the same terms of license than the past code. Was the last code taken from anywhere or totally created here?

This is exactly what needs to be figured out as part of this PR. Find out where the older code comes from, whether we can use it or not, document it and explain how to update and so on. You can't just add whatever license we have for the general project to external code.

vdelacruzb:
Did some refactoring and avoided using totally the old code due to possible license uses.
Also removed tilecover.js as it was a browserified version of @mapbox/tilebelt and @mapbox/tile-cover with some of our code. Now the makefile directly browserify the quadint code with these two modules which already contain auxiliar functions to make it work.
I added the LICENSE of the modules to the js folder.

quadkey/bq/Makefile
```diff
@@ -0,0 +1,77 @@
...
+
+dataset_remove: check_environment
+	${BQ} rm -r -f -d $(BQ_PROJECTID):$(BQ_DATASET_QUADKEY)
+
+REPLACEMENTS = 	-e 's!@@BQ_PROJECTID@@!$(BQ_PROJECTID)!g' \
+		-e 's!@@BQ_DATASET_QUADKEY@@!$(BQ_DATASET_QUADKEY)!g' \
+		-e 's!@@QUADKEY_BQ_LIBRARY@@!$(QUADKEY_BQ_LIBRARY)!g'
+
+dataset_deploy: check_environment
+	for n in $(wildcard sql/*.sql); do \
```
Algunenano, line 57:
```suggestion
	for n in $(sort $(wildcard sql/*.sql)); do \
```

quadkey/bq/sql/ST_ASQUADINTPOLYFILL.sql
```diff
@@ -0,0 +1,28 @@
+-----------------------------------------------------------------------
+--
+-- Copyright (C) 2021 CARTO
+--
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_QUADKEY@@.POLYFILL_FROMGEOJSON`
```
Algunenano, line 7: I understand that this function is private right? In that case I would name it as __ST_ASQUADINT_POLYFILL

quadkey/bq/sql/ST_ASQUADINTPOLYFILL.sql
```diff
@@ -0,0 +1,28 @@
...
+    {
+        throw new Error('NULL argument passed to UDF');
+    }
+
+    let pol = JSON.parse(geojson);
+    let quadints = geojsonToQuadints(pol, {min_zoom: resolution, max_zoom: resolution});
+    return quadints.map(String);
+""";
+
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_QUADKEY@@.ST_ASQUADINTPOLYFILL`
```
Algunenano, line 24:
```suggestion
CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_QUADKEY@@.ST_ASQUADINT_POLYFILL`
```

For consistency with `ST_ASH3_POLYFILL`

quadkey/bq/sql/ST_GEOGFROMQUADINT_BOUNDARY.sql
```diff
@@ -0,0 +1,27 @@
+-----------------------------------------------------------------------
+--
+-- Copyright (C) 2021 CARTO
+--
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_QUADKEY@@.GEOJSONBOUNDARY_FROMQUADINT`
```
Algunenano, line 7: Same, this is private, so I would call it as the main function and add the `__` prefix.

quadkey/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,148 @@
...
+
+{{% bannerNote type="code" %}}
+quadkey.QUADKEY_FROMQUADINT(quadint INT64) -> STRING
+{{%/ bannerNote %}}
+
+* `quadint`: `INT64` quadint we want to convert to quadkey.
+
+Transform a quadint index to an equivalent quadkey.
+
+#### quadkey.PARENT
```
Algunenano, line 67: In H3 this is called `TOPARENT` and supports passing the desired resolution. Thoughts?

quadkey/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,148 @@
...
+
+{{% bannerNote type="code" %}}
+quadkey.PARENT(quadint INT64) -> INT64
+{{%/ bannerNote %}}
+
+* `quadint`: `INT64` quadint we want to get the parent from.
+
+Returns the parent quadint of a given quadint. A parent quadint is a quadint of smaller level of detail (zoom - 1) which contains the current quadint.
+
+#### quadkey.CHILDREN
```
Algunenano, line 77: Same, in `h3` this function is called `TOCHILDREN` and supports passing the resolution of the children you want.

quadkey/library/js/quadint.js
```diff
@@ -0,0 +1,190 @@
...
+    }
+    if (direction === 'right') {
+        x = x < tilesPerLevel - 1 ? x + 1 : 0;
+    }
+    if (direction === 'up') {
+        y = y > 0 ? y - 1 : tilesPerLevel - 1;
+    }
+    if (direction === 'down') {
+        y = y < tilesPerLevel - 1 ? y + 1 : 0;
+    }
```
Algunenano, line 144: I think checking the direction is one of the valid values here would be best, otherwise you get the same input and no error (e.g. "Left" won't work as expected).

quadkey/library/test/quadkey_test.js
```diff
@@ -0,0 +1,140 @@
...
+                zxyDecoded = ZXYFromQuadint(quadintFromZXY(z, x, y));
+                zDecoded = zxyDecoded.z;
+                xDecoded = zxyDecoded.x;
+                yDecoded = zxyDecoded.y;
+                assert.ok(z === zDecoded && x === xDecoded && y === yDecoded);
+            }
+        }
+    });
+
+    it('Should be able to encode/decode between quadint and quadkey at any level of zoom', async() => {
```
Algunenano, line 48: I think I would add some known value tests and not only rountrip, otherwise you risk breaking compatibility and not noticing it.

quadkey/bq/test/quadkey_integration.js
```diff
@@ -0,0 +1,500 @@
+const assert = require('assert').strict;
+const {BigQuery} = require('@google-cloud/bigquery');
+
+const BQ_PROJECTID = process.env.BQ_PROJECTID;
+const BQ_DATASET_QUADKEY = process.env.BQ_DATASET_QUADKEY;
+
+describe('QUADKEY integration tests', () => {
```
Algunenano, line 7: Personally I prefer to break down the integration tests in multiple files, it's clearer to see what's being tested in each of them and they are executed in parallel (so everything is much faster). In any case, this looks good.

Algunenano (commented):
Looks great. I've added some comments here and there about possible improvements, mostly to keep things consistent now that we have more modules added (so it could be changed in the other modules instead).

quadkey/bq/doc/REFERENCE.md
vdelacruzb: This should be doable on my side. I go for it.

quadkey/bq/doc/REFERENCE.md
vdelacruzb: This should be also doable. I guess this should return all the children contained for a given resolution right? Example, current quadint is in resolution 1, TOCHILDREN(quadint, 3)-> array with 16 children.

quadkey/bq/test/quadkey_integration.js
vdelacruzb: I will split it then as it should be a fast change.

quadkey/library/js/quadint.js
vdelacruzb: oki doki. I'm going to check for expected values. Do you like the idea of doing also a toLowerCase? or should be less flexible and not allow capital letters?

Algunenano: Doing `toLowerCase` seems llike a good idea

Algunenano (approved):
Looks good

PR #32, opened 2021-02-17
Sphericalmercator module

vdelacruzb:


ibrahimmenem (approved):
:+1: for ci.yml

sphericalmercator/README.md
```diff
@@ -0,0 +1,6 @@
+# Spherical Mercator
```
Algunenano, line 1: It would nice to have a description about what it is, where it comes from...

sphericalmercator/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,28 @@
+## Reference
+
+### SPHERICALMERCATOR
+
+#### BBOX
+
+{{% bannerNote type="code" %}}
```
Algunenano, line 7: Please don't use `NUMERIC`. Also, I have no idea what this is supposed to be returning from the description so that's not a good signal.

sphericalmercator/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,28 @@
...
+{{%/ bannerNote %}}
+
+Returns bbox array of values in form [w, s, e, n].
+
+* `x`: `NUMERIC` x (longitude) number.
+* `y`: `NUMERIC` y (latitude) number.
+* `zoom`: `NUMERIC` zoom level.
+* `tileSize`: `NUMERIC` size of the tiles we want to use.
+
+#### XYZ
```
Algunenano, line 18: Are we using this (or the other function). How is this different from quadkey or what you can achieve with quadkeys? Looks pretty similar to me.

Algunenano (changes requested):
What's the difference between this and quadkeys? What does this library offers that isn't covered by the quadkey module and why isn't it better to add it there?

Jesus89:
Won't be included for now

PR #34, opened 2021-02-18
S2 module

vdelacruzb:


s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+
+* `id`: `INT64` s2 cell id.
+
+#### LONG_FROMKEY
+
+{{% bannerNote type="code" %}}
+s2.LONG_FROMKEY(key STRING)
+{{%/ bannerNote %}}
+
+Extract the longitude component from a quadkey.
```
SpencerC, line 95: Suggest using the parlance "Hilbert quadkey" or something here to differentiate from the quadkey module.

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+* `point`: `GEOGRAPHY` point we want to get the quadint from.
+* `level`: `NUMERIC` Level of detail or zoom.
+
+#### ST_GEOGFROMLATLNG_BOUNDARY
+
+{{% bannerNote type="code" %}}
+s2.ST_GEOGFROMLATLNG_BOUNDARY(latitude FLOAT64, longitude FLOAT64, level NUMERIC)
+{{%/ bannerNote %}}
+
+Returns the geography boundary for a given level of detail and geographic coordinates. We extract the boundary by getting the corner longitudes and latitudes, then enclose it in a GEOJSON and finally transform it to geography.
```
SpencerC, line 148: GEOJSON -> GeoJSON

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+
+{{% bannerNote type="code" %}}
+s2.LAT_FROMKEY(key STRING)
+{{%/ bannerNote %}}
+
+Extract the latitude component from a quadkey.
+
+* `key`: `STRING` quadkey.
+
+#### LONGLAT_ASS2
```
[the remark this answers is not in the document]
vdelacruzb: Yes, I'm not super fan either of the double S but where using the same notation in all the other modules LONGLAT_ASH3, LONGLAT_ASQUADINT etc..
We rarely use more than one underscore in the names as well, mainly when it's a ST_XXXX_POINT/BOUNDARY.

What about, S2_FROMLONGLAT?

s2/bq/doc/REFERENCE.md
vdelacruzb: Actually in the quadkey module we are using majorly, quadints, which is a quadkey stored in a INT64. But yes, using HILBERTQUADKEY in this module should make it more clear.

vdelacruzb:
What do you think of having LONGLAT_FROMID and LONGLAT_FROMKEY instead of LONG_FROMID + LATFROMID?
It would return an array of float with 2 values or we could make it return an struct<long, lat>. 
Maybe it's more comfortable for users.

vdelacruzb:
Fixed typos, still on discussion about functions naming

s2/bq/doc/REFERENCE.md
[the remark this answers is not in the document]
SpencerC: Something like `S2_FROMLONGLAT` SGTM. You might consider using `ID_FROMLONGLAT` or `LONGLAT_ASID` since in practice these calls will be prefixed with `<package>.s2.` (right?).

s2/bq/doc/REFERENCE.md
SpencerC: Just to make sure: my suggestion just pertained to documentation. The verbosity of `LONG_FROMHILBERTQUADKEY` might negatively impact function readability more than the inline context might help it, especially since this is a function developers are likely to use as an argument to other functions, ex.
`ST_GEOGPOINT(LONG_FROMHILBERTQUADKEY(columnY), LAT_FROMHILBERTQUADKEY(columnX))`

SpencerC:
> What do you think of having LONGLAT_FROMID and LONGLAT_FROMKEY instead of LONG_FROMID + LATFROMID?

Maybe all of the above? `STRUCT`s and `ARRAY`s can slightly complicate querying and might be less appropriate for certain applications than lat/long functions that can put results straight into columns. One nice shortcut might be to provide a function that returns an `ST_GEOGPOINT` from a provided ID or key.

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+* `latitude`: `FLOAT64` vertical coordinate on the map.
+* `level`: `NUMERIC` Level of detail or zoom.
+
+#### LONG_FROMID
+
+{{% bannerNote type="code" %}}
+s2.LONG_FROMID(id INT64)
+{{%/ bannerNote %}}
+
+Extract the longitude component from a give s2 cell id.
```
SpencerC, line 85: Suggest throwing in the word 'centroid' here. Obvious to experienced folks, maybe not to everyone.

s2/library/js/s2geometry.js
```diff
@@ -0,0 +1,1924 @@
...
+//NOTE: compared to the google S2 geometry library, we vary from their code in the following ways
+// - cell IDs: they combine face and the hilbert curve position into a single 64 bit number. this gives efficient space
+//             and speed. javascript doesn't have appropriate data types, and speed is not cricical, so we use
+//             as [face,[bitpair,bitpair,...]] instead
+// - i,j: they always use 30 bits, adjusting as needed. we use 0 to (1<<level)-1 instead
+//        (so GetSizeIJ for a cell is always 1)
+
+
+
+/**
```
Algunenano, line 27: Yeah, no no no no no. I don't like this at all. First, BigQuery has WASM, second we have tools to generate this from C or C++ code and not include a blob here; this needs to go away.

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+{{%/ bannerNote %}}
+
+Returns a GeoJSON containing the boundary corners of a quadkey.
+
+* `key`: `STRING` quadkey to extract the corners from.
+
+#### GEOJSONBOUNDARY_FROMLONGLAT
+
+{{% bannerNote type="code" %}}
+s2.GEOJSONBOUNDARY_FROMLONGLAT(longitude FLOAT64, latitude FLOAT64, level NUMERIC)
```
Algunenano, line 26: Let's remove `NUMERIC` from the face of Earth, please.

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+{{% bannerNote type="code" %}}
+s2.VERSION()
+{{%/ bannerNote %}}
+
+Returns the current version of the s2 library.
+
+#### GEOJSONBOUNDARY_FROMKEY
+
+{{% bannerNote type="code" %}}
+s2.GEOJSONBOUNDARY_FROMKEY(key STRING)
```
Algunenano, line 16: Do we need geojson from the open API?

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+
+Returns the current version of the s2 library.
+
+#### GEOJSONBOUNDARY_FROMKEY
+
+{{% bannerNote type="code" %}}
+s2.GEOJSONBOUNDARY_FROMKEY(key STRING) 
+{{%/ bannerNote %}}
+
+Returns a GeoJSON containing the boundary corners of a quadkey.
```
Algunenano, line 19: `quadkey` -> `s2 cell`?

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+
+* `key`: `STRING` quadkey to extract the corners from.
+
+#### GEOJSONBOUNDARY_FROMLONGLAT
+
+{{% bannerNote type="code" %}}
+s2.GEOJSONBOUNDARY_FROMLONGLAT(longitude FLOAT64, latitude FLOAT64, level NUMERIC)
+{{%/ bannerNote %}}
+
+Returns a GeoJSON containing the boundary corners given a longitude, latitude and zoom level.
```
Algunenano, line 29: Do we really need this function? We should have `long, lat, level -> id`  and `id -> geog (boundary)`. I don't see why we want to maintain `long, lat, level -> geojson boundary`. This isn't meant to match a JS API, but to provide a good SQL API we can build with.

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+
+{{% bannerNote type="code" %}}
+s2.KEY_FROMID(id INT64)
+{{%/ bannerNote %}}
+
+Convert from s2 cell id to hilbert quadtree id.
+
+* `id`: `INT64` s2 cell id to be converted.
+
+#### KEY_FROMLONGLAT
```
Algunenano, line 67: I'm having issues understanding what's a `key` and what's an `id`. What's the deal with it? Can't we use the 64-bit [cell id](https://s2geometry.io/devguide/s2cell_hierarchy.html#s2cellid-numbering)? That should uniquely identify the cell ids and we can use that binary representation in an INT64 right?

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+s2.KEY_FROMLONGLAT(longitude FLOAT64, latitude FLOAT64, level NUMERIC)
+{{%/ bannerNote %}}
+
+Generate the quadkey for a given longitude, latitude and zoom level.
+
+* `longitude`: `FLOAT64` horizontal coordinate on the map.
+* `latitude`: `FLOAT64` vertical coordinate on the map.
+* `level`: `NUMERIC` Level of detail or zoom.
+
+#### LONG_FROMID
```
Algunenano, line 79: What's the reasoning for this function? As @SpencerC mentions, the fact that this is returning the latitude of the centroid is far from obvious and I'm not sure if we want to provide that.

A good start would be to have a function that returns the geography that represents the cell, then if people want the centroid or the latitude of the centroid they can get it using BigQuery functions. If we saw that everybody was using this for their visualizations (God forbit them) then we could think of adding a helper function (again, using the battle tested BigQuery functions), but as far as I can see currently it doesn't make much sense to provide this.

s2/bq/doc/REFERENCE.md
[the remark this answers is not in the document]
Algunenano: Yep, naming is troublesome here. We wanted to be consistent with BigQuery names so the stardard would be `LONGLAT_ASS2` and `ST_ASS2` which isn't great. We can absolutely change it (and probably should) but we should be consistent and also follow whatever standard we define here for the other modules too

s2/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,162 @@
...
+{{% bannerNote type="code" %}}
+s2.ST_ASS2(point GEOGRAPHY, resolution NUMERIC)
+{{%/ bannerNote %}}
+
+Converts a given point at given level of detail to a s2 cell id.
+
+* `point`: `GEOGRAPHY` point we want to get the quadint from.
+* `level`: `NUMERIC` Level of detail or zoom.
+
+#### ST_GEOGFROMLONGLAT_BOUNDARY
```
Algunenano, line 142: Again, lots of duplicated functionality here. We should start simple and build from there, this just adds a lot of complexity to something that should be as simple as I have a geography and I want the s2 cell id, or I have the s2 cell id and I want the geography.

s2/library/js/s2geometry.js
```diff
@@ -0,0 +1,1924 @@
+/// S2 Geometry functions
```
Algunenano, line 1: This looks like a modified version from https://git.coolaj86.com/coolaj86/s2-geometry.js/src/branch/master/src/s2geometry.js and I'm not too happy with the approximation (it's neither well supported upstream nor here). How hard it would be to cover the basic use cases with the official C++ library (https://github.com/google/s2geometry)?

Algunenano (changes requested):

s2/bq/doc/REFERENCE.md
vdelacruzb: well, I would say this is not intended to be used directly but from ST_GEOGFROMKEY_BOUNDARY in order to get the the geography. So I will remove it from the Reference

s2/bq/doc/REFERENCE.md
vdelacruzb: With key we mean HILBERTQUADKEY, example, '4/032212303102210' and with id we mean the S2 cell ID which is an INT64 as you say. Yes, both can uniquely identify the cell, so to simplify the module I think it's a good idea to do what you say. This is a similar situation to our quadkey/quadint module, so I will keep also the parsers as we do with the quadkey module.

vdelacruzb:
I added your changes and tried to make this module way simpler.

Now everything is handled with IDs and I just keep a couple of functions to convert to/from Hilbert Quadkey. I think it's interesting to use the INT64 as identifier even if internally is doing some conversions to/from quadkey.

I have pending still to create the wasm directly from Google s2 C++ library, which can take some more time so I have created a ticket for it, https://app.clubhouse.io/cartoteam/story/140132/use-official-google-library-for-s2.

s2/bq/Makefile
```diff
@@ -0,0 +1,77 @@
...
+
+dataset_remove: check_environment
+	${BQ} rm -r -f -d $(BQ_PROJECTID):$(BQ_DATASET_S2)
+
+REPLACEMENTS = 	-e 's!@@BQ_PROJECTID@@!$(BQ_PROJECTID)!g' \
+		-e 's!@@BQ_DATASET_S2@@!$(BQ_DATASET_S2)!g' \
+		-e 's!@@S2_BQ_LIBRARY@@!$(S2_BQ_LIBRARY)!g'
+
+dataset_deploy: check_environment
+	for n in $(wildcard sql/*.sql); do \
```
Algunenano, line 57:
```suggestion
	for n in $(sort $(wildcard sql/*.sql)); do \
```

s2/bq/sql/VERSION.sql
```diff
@@ -0,0 +1,14 @@
+-----------------------------------------------------------------------
+--
+-- Copyright (C) 2021 CARTO
+--
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_S2@@.VERSION`()
+    RETURNS INT64
```
Algunenano, line 8: Since this is tied up to a `s2-geometry.js` release I would rather return a string with the release of that library. Personal preference.

s2/bq/test/s2_integration.js
```diff
@@ -0,0 +1,141 @@
...
+        let rows;
+        await assert.doesNotReject( async () => {
+            const [job] = await client.createQueryJob({ query: query });
+            [rows] = await job.getQueryResults();
+        });
+        assert.equal(rows.length, 1);
+        assert.equal(rows[0].versioncol, 1);
+    });
+
+    it ('KEY / ID conversions should work', async () => {
```
Algunenano, line 31: Testing compatiblity is great, but this should also check static values to confirm they are correct.  Also it should include a test for https://github.com/CartoDB/carto-spatial-extension/issues/29

s2/bq/sql/ST_GEOGFROMID_BOUNDARY.sql
```diff
@@ -0,0 +1,37 @@
+-----------------------------------------------------------------------
+--
+-- Copyright (C) 2021 CARTO
+--
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_S2@@.GEOJSONBOUNDARY_FROMID`
```
Algunenano, line 7: This should be private IMO

s2/bq/test/s2_integration.js
```diff
@@ -0,0 +1,141 @@
...
+            "coordinates": [[
+                [124.99991607494462, -14.000016145055083],
+                [124.99991607494462, -13.99970528488021],
+                [125.0002604046465, -13.999648690569117],
+                [125.0002604046465, -13.999959549588995],
+                [124.99991607494462, -14.000016145055083]
+            ]]
+        };
+
+        let query = `SELECT \`${BQ_PROJECTID}\`.\`${BQ_DATASET_S2}\`.GEOJSONBOUNDARY_FROMID(
```
Algunenano, line 82: I would test ST_GEOGFROMID_BOUNDARY instead since that's the public API, and use other function from BigQuery if necessary to print that geography.

s2/library/test/s2_test.js
```diff
@@ -0,0 +1,39 @@
...
+/* Emulate how BigQuery would load the file */
+global.eval(fs.readFileSync('../../s2_library.js') + '');
+
+describe('S2 unit tests', () => {
+    it('Version', async() => {
+        assert.equal(s2Version(), 1);
+    });
+
+    it('Quadkey/S2 id Conversions', async() => {
+        let z, lat, lng;
```
Algunenano, line 12: Same as in other cases, I think it would be great to have tests where you check specific values and match to the known good value, so not everything checks that a -> b -> a works.

Algunenano (commented):
Great job. I've left some comments (similar to the ones in quadkey).

Algunenano (approved):
LGTM. As I mentioned before, it would be great to have tests covering exact values (and not only round trip) but those could be added in the future. Please rebase on top of master and pass CI before merging.

PR #35, opened 2021-02-18
Turf module

vdelacruzb:


turf/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,96 @@
...
+* `geojson`: `STRING` input to be buffered.
+* `radius`: `NUMERIC` distance to draw the buffer (negative values are allowed).
+* `units`: `STRING` any of the options supported by turf units: miles, kilometers, and degrees.
+* `steps`: `NUMERIC` number of steps.
+
+``` sql
+SELECT jslibs.turf.ST_BUFFER(ST_GEOGPOINT(-74.00,40.7128),1,'kilometers',10) as geo
+```
+
+#### ST_SIMPLIFY
```
Algunenano, line 86: BigQuery already has a [ST_Simplify](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_simplify), do we really need another (same for SIMPLIFY)?

turf/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,96 @@
...
+
+{{% bannerNote type="code" %}}
+turf.VERSION()
+{{%/ bannerNote %}}
+
+Returns the current version of the turf library.
+
+#### BBOX
+
+{{% bannerNote type="code" %}}
```
Algunenano, line 27: Why are we exposing this? Why are we asking users to do turf.BBOX(ST_AsGeoJSON(g)) instead of having a ST_BBOX or similar. Also, how is this different from `tiler.__WKB_ENVELOPE`?

turf/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,96 @@
...
+
+Takes a set of features, calculates the bbox of all input features, and returns a bounding box (minX, minY, maxX, maxY).
+https://turfjs.org/docs/#bbox
+
+* `geojson`: `STRING` geojson to extract the bounding box from.
+
+#### BUFFER
+
+{{% bannerNote type="code" %}}
+turf.BUFFER(geojson STRING,radius NUMERIC, options STRUCT<unit STRING,steps NUMERIC>)
```
Algunenano, line 39: Instead of `NUMERIC` which doesn't make sense to interface with JS, let's use `INT64` or `FLOAT64` or any numeric type more appropiate to this situation.

Algunenano, line 39: What's with the `STRUCT` here just to support a single option? It's not in ST_BUFFER

turf/README.md
```diff
@@ -0,0 +1,10 @@
+# Turf
+
+Taken from here https://turfjs.org
```
Algunenano, line 3: Again, where does the external code comes from, what release, how to update it...?

turf/bq/test/turf_integration.js
```diff
@@ -0,0 +1,118 @@
...
+        let rows;
+        await assert.doesNotReject( async () => {
+            const [job] = await client.createQueryJob({ query: query });
+            [rows] = await job.getQueryResults();
+        });
+        assert.equal(rows.length, 1);
+        assert.equal(rows[0].versioncol, 1);
+    });
+
+    it ('BBOX should work', async () => {
```
Algunenano, line 31: We should be exposing and testing things with geographies and forget/remove the GeoJSON API. Also, in general (not only this specific function) this is missing checking possible errors (NULL, invalid parameters, etc).

Algunenano (changes requested):

Jesus89:
Will not be included for now

PR #36, opened 2021-02-19
Placekey module

vdelacruzb:


vdelacruzb:
Not fan of LATLONG_FROMPLACEKEY returning a string “lat,long”.
Why not an array of two floats or a struct [long, lat] LONGLAT_FROMPLACEKEY.
What do you think?

vdelacruzb:
I'm iterating on this PR as there are some issues with CI checks

placekey/library/js/h3-js.umd.js
```diff
@@ -0,0 +1,2 @@
...if(!r||"number"!=typeof r.i||"number"!=typeof r.j)throw new Error("Coordinates must be provided as an {i, j} object");var i,o,a,u=e._malloc(w),l=e._malloc(f);a=(o=r).j,e.setValue(i=u,o.i,"i32"),e.setValue(i+t,a,"i32");var c=n.experimentalLocalIjToH3.apply(n,b(A).concat([u],[l])),s=v(l);if(e._free(u),e._free(l),0!==c)throw new Error("Index not defined for this origin and IJ coordinates pair. IJ coordinates may be too far from origin, or a pentagon distortion was encountered.");return s},A.hexArea=function(A,e){switch(d(A),e){case g.m2:return n.hexAreaM2(A);case g.km2:return n.hexAreaKm2(A);default:throw new Error("Unknown unit: "+e)}},A.edgeLength=function(A,e){switch(d(A),e){case g.m:return n.edgeLengthM(A);case g.km:return n.edgeLengthKm(A);default:throw new Error("Unknown unit: "+e)}},A.numHexagons=function(A){d(A);var e=E(n.numHexagons(A)),r=e[0],i=e[1];return i?i*Math.pow(2,32)+r:r},A.getRes0Indexes=function(){var A=n.res0IndexCount(),r=e._malloc(f*A);n.getRes0Indexes(r);var i=I(r,A);return e._free(r),i},A.getPentagonIndexes=function(A){d(A);var r=n.pentagonIndexCount(),i=e._malloc(f*r);n.getPentagonIndexes(A,i);var t=I(i,r);return e._free(i),t},A.degsToRads=G,A.radsToDegs=x});
```
Algunenano, line 1: This is already being imported by the H3 module. It should depend on it, and not duplicate an extra library.

placekey/README.md
```diff
@@ -0,0 +1,8 @@
+# Placekey
+
+Based on https://github.com/Placekey/placekey-js
```
Algunenano, line 3: Please document the release and the changes done to the external code

placekey/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### VERSION
+
+{{% bannerNote type="code" %}}
+placekey.VERSION()
+{{%/ bannerNote %}}
+
+Returns the current version of the placekey library.
+
+#### PLACEKEY_FROMH3
```
Algunenano, line 13: I'm not sure about this `FROMH3` and `H3_FROMPLACEKEY` reinventing the wheel and changing how we work with h3 indexes (using string instead of int64 and so on).

placekey/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,72 @@
...
+
+{{% bannerNote type="code" %}}
+placekey.PLACEKEY_FROMH3(h3Index STRING)
+{{%/ bannerNote %}}
+
+Transform a h3 index to an equivalent placekey.
+
+* `h3Index`: `STRING` h3 index we want to convert to placekey.
+
```
Algunenano, line 22: Missing docs about the return values

placekey/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,72 @@
...
+{{%/ bannerNote %}}
+
+Return the geographic coordinates corresponding to a given placekey. The return format is a STRING containing the longitude and latitude separated by a comma.
+
+* `placekey`: `STRING` placekey we want to extract the coordinates from.
+
+#### ST_GEOGFROMPLACEKEY_POINT
+
+{{% bannerNote type="code" %}}
+placekey.ST_GEOGFROMPLACEKEY_POINT(placekey STRING)
```
Algunenano, line 67: I don't understand why the `_POINT` part

Algunenano (changes requested):
I have lots of mixed feelings about this. Considering the where part of a placekey is literally a different way of "naming" h3 indexes, maybe we should limit ourselves to do the conversion between placekey and h3 int index and leave the rest to the h3 module (we can have wrappers if we really need them).

Let's hold this PR until the h3 module is in place.

placekey/library/js/h3-js.umd.js
vdelacruzb: Initially I implemented it like that. Started to have issues with the ci. At some point I though that maybe we wanted totally independent modules + I was having issues with the CI, so I grouped everything together. In some way I like it depending on the other module, so we don't have duplicated codes to maintain.

Algunenano:
Replaced by #41

PR #37, opened 2021-03-02
Improve CI

Algunenano:
Applies several fixes that appear in multiple PRs and some extra necessary to improve things overall:
* Avoid using datasets starting with numbers (they are fragile in BigQuery, as they work in some situations but not in others).
* Use node 14.
* Use query() instead of createQueryJob(). Query() has only one point of failure, while createQueryJob() and friends might fail in different functions for the exact same query (specially with procedures).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ on:
 
 env:
   GCLOUD_VERSION: 'latest'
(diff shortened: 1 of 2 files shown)


vdelacruzb (approved):
Nice, go for it.

jvillarf (commented):
:ok_hand:

Issue #29, opened 2021-02-12
ST_S2 returning crazy values


SpencerC:
```
WITH data AS (SELECT jslibs.s2.ST_S2(ST_GEOGPOINT(-74.006, 40.7128), 12) AS id)
SELECT * FROM data
```
Returns
```
Failed to coerce output value "9926595690882924544" to type INT64: Bad int64 value: 9926595690882924544
```
This seems to be the case for geopoints with longitudes between -60 and -120 and latitudes below about (but not exactly) 40.

francois-baptiste:
Thank you @SpencerC for raising this issue.
Looks like it's a problem of casting uint64 into int64.
The JS UDF should output directly int64 instead of uint64.
I will digg into it.

SpencerC:
Since S2 cell numbers are unsigned, wouldn't it be better to solve the underlying issue and keep the return type `UINT64`?

francois-baptiste:
Sadly enough `uint64` is not a valid type in bigquery.

> We disallow uint, uint64 and uintptr, because they are (potentially) 64-bit unsigned values, so they cannot be accurately represented as a signed 64-bit integer.

That's why I propose to use `int64` with the same binary representation as a  `uint64`.

francois-baptiste:
fixed by #38

francois-baptiste:
Sorry, I just realized that I kept this fix for myself a long time ago and forgot to publish the patch in PR.
Since @vdelacruzb  is currently working on an entire refactor of the project, I PR directly the refactored version.

You can preview the fix as below using my personal GCS bucket:

```
CREATE TEMP FUNCTION latLngToId(latitude FLOAT64, longitude FLOAT64, level NUMERIC) RETURNS INT64 DETERMINISTIC LANGUAGE js
OPTIONS (library=["gs://bigquery-geolib/s2geometry.js"]) AS """
return S2.keyToId(S2.S2Cell.latLngToKey(latitude, longitude, level));
""";
SELECT latLngToId(-74.006, 40.7128, 12)

```

francois-baptiste:
@SpencerC see #39 if you want to output your S2 cells to kepler.gl

Algunenano:
@vdelacruzb Can you confirm if this is fixed and tested with the new code, please?

vdelacruzb:
Yes, this has being tested. And also have a specific integration test entry for this case.

PR #38, opened 2021-03-02
S2.keyToId return INT64 instead of UINT64

francois-baptiste:
S2.keyToId return INT64 instead of UINT64 but with the same binary representation
This is because UINT64 is not a valid type in bigquery.

### DIFF
--- a/s2/library/js/s2geometry.js
+++ b/s2/library/js/s2geometry.js
@@ -1797,7 +1797,7 @@ var S2 =
         bin += '0';
       }
 
-      return Long.fromString(bin, true, 2).toString(10);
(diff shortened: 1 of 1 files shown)


francois-baptiste:
This fix #29

Algunenano:
cc/ @vdelacruzb. Please take this into account once you come back to the S2 module.

vdelacruzb:
Thanks a lot @francois-baptiste I was struggling with this same thing yesterday afternoon, I was missing the bit of the sign.
I close this PR as I just introduced your change in https://github.com/CartoDB/carto-spatial-extension/pull/34.

PR #40, opened 2021-03-04
H3 module

Algunenano:
What this does:
* Bring the H3 module to the extension.
* Includes a script to download and keep h3.min.js synchronized if necessary.
* BigQuery integration and tests.
* Reference docs

What this doesn't do:
* Expose all the functionality of h3-js. There are functions that I haven't brought in yet because they have a good BigQuery alternative (like getting the centroid or area of a cell, or distance in meters between cells) and some because I'm not sure how useful they are but they could be added easily if necessary (like h3Line). Let me know if you see anything missing and I'll add it.


h3/bq/sql/AS_H3.sql
```diff
@@ -0,0 +1,68 @@
...
+    LANGUAGE js
+    OPTIONS (library=["@@H3_BQ_LIBRARY@@"])
+AS
+"""
+    if (!geojson || _resolution == null) {
+        return null;
+    }
+
+    const resolution = Number(_resolution);
+    if (resolution < 0 || resolution > 15) {
```
vdelacruzb, line 43: In this kind of cases what do we want, to return nothing or throwing an error. I was throwing some errors in quadkey for wrong zooms since I thought it would be easier to track for the user.

Now that I see this, I will also include some more checks for null arguments.

h3/bq/sql/H3_DISTANCE.sql
```diff
@@ -0,0 +1,33 @@
...
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_H3@@.__H3_DISTANCE`(index_lower1 INT64, index_upper1 INT64, index_lower2 INT64, index_upper2 INT64)
+    RETURNS INT64
+    DETERMINISTIC
+    LANGUAGE js
+    OPTIONS (library=["@@H3_BQ_LIBRARY@@"])
+AS
+"""
+    if (index_lower1 == null || index_upper1 == null || index_lower2 == null || index_upper2 == null)
+        return null;
+    const index1 = [Number(index_lower1), Number(index_upper1)];
```
vdelacruzb, line 16: Some functions are checking if H3 inputs are valid and some not.
Maybe add here the check for both inputs.

h3/bq/sql/H3_ISPENTAGON.sql
```diff
@@ -0,0 +1,24 @@
...
+
+CREATE OR REPLACE FUNCTION `@@BQ_PROJECTID@@.@@BQ_DATASET_H3@@.__H3_ISPENTAGON`(index_lower INT64, index_upper INT64)
+    RETURNS BOOLEAN
+    DETERMINISTIC
+    LANGUAGE js
+    OPTIONS (library=["@@H3_BQ_LIBRARY@@"])
+AS
+"""
+    if (index_lower == null || index_upper == null)
+        return false;
```
vdelacruzb, line 15: Here we are not checking if is valid neither.

h3/bq/sql/H3_DISTANCE.sql
Algunenano: This is because internally the module is already handling it. There are integration tests checking it just in case.

h3/bq/sql/H3_ISPENTAGON.sql
Algunenano: Same, and it's covered by the integration tests.

h3/bq/sql/AS_H3.sql
Algunenano: I chose returning `NULL` because the underlying library was already doing that in most cases, and in SQL is a huge pain when you run a function over millions of records and it's stopped because a single failure. BigQuery has `SAFE.` but UDFs don't have that option. In any case it could be changed in the future to something different.

vdelacruzb (approved):
Nicely done. Ton of tests and very well documented.

PR #41, opened 2021-03-04
Placekey module on top of h3

Algunenano:
Simplify the initial proposal of the placekey module to reduce it to placekey -> h3, h3->placekey and validation. Any other operations (like geotoplacekey and so on) where purely based on h3 itself, so the recommended way for now is to transform things to h3 and work with that. In the future we could consider adding wrappers that make use of the existing placekey or h3 module.

Builds on top of the H3 module, so it should wait until that one is merged so it can be rebased and cleaned up.

Closes https://github.com/CartoDB/carto-spatial-extension/pull/36 since it expands what work was done there.


placekey/library/test/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "main": "placekey_test.js",
+  "scripts": {
+    "test": "mocha"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "h3-js": "^3.7.0",
+    "mocha": "^8.2.1",
```
vdelacruzb, line 14: do we still need mocha?

vdelacruzb (approved):
LGTM.
I don't think the mocha package is needed in library/test so please remove it before merging.

placekey/library/test/package.json
Algunenano: You are right. I'll remove it

PR #42, opened 2021-03-05
Remove prefix

Algunenano:
As we've talked, I've removed the `H3_` prefix from the function names. I've only left the `H3_ASHEX` and `H3_FROMHEX`, but those could be changed to `h3.ASHEX` and `h3.FROMHEX` if we wanted to.

I've also added sorting to the deployment so it's consistent and avoids possible dependency issues.


vdelacruzb (approved):
Nice, go for it.

PR #43, opened 2021-03-08
Tiler quadkey support

vdelacruzb:


Algunenano (changes requested):
Code looks good, but I don't like the name (it doesn't reflect what is doing and references a different module altogether).

vdelacruzb:
Renamed the function to LONGLAT_ASQUADINTLIST. Normally we are using LONGLAT_ for functions that accepts map coordinates.  Was thinking of a word that could sum up "Per zoom level", any suggestion?

vdelacruzb:
There was a bug in quadintFromLocation sending the wrong zoom level to get a quadint. This was causing the tiler to not create the tilesets properly.
Also reduced the latitude limits on tests, since quadintFromLocation get messed when we get too close to the poles.

quadkey/bq/doc/REFERENCE.md
```diff
@@ -147,4 +147,19 @@ quadkey.ST_GEOGFROMQUADINT_BOUNDARY(quadint INT64) -> GEOGRAPHY
 
 * `quadint`: `INT64` quadint we want to get the boundary geography from.
 
-Returns the geography boundary for a given quadint. We extract the boundary the same way as when we calculate the bbox, then enclose it in a GEOJSON and finally transform it to geography.
\ No newline at end of file
+Returns the geography boundary for a given quadint. We extract the boundary the same way as when we calculate the bbox, then enclose it in a GEOJSON and finally transform it to geography.
+
+### quadkey.LONGLAT_ASQUADINTLIST
```
Algunenano, line 152: I would add `_RESOLUTION` or something like that here, since you are calculating a subcell of the tile (based on the resolution).

Algunenano (changes requested):
> Also reduced the latitude limits on tests, since quadintFromLocation get messed when we get too close to the poles.

Detecting this was great, but the solution you applied is the opposite of great. I don't think we want to remove the tests, instead we should make sure the code handles it. For example, the tiler clips the latitude to ± 86, but we could be more conservative and use the exact 85.06 defined for webmercator (https://epsg.io/3857)

Algunenano:
Keep in mind that users of BigQuery (or any other geospatial database) can, and likely many will, have data around the poles, so crashing or returning rubish isn't great. I prefer clipping to the nearest representable quadkey/webmercator point in those situations.

Algunenano (approved):
:+1:

PR #44, opened 2021-03-10
update kring function

vdelacruzb:
The function KRING should accept different distances.
*Notice that the test values where just reordered for the new implementation, but the values contained are the same. It's just that deepEqual complains if the order is not exactly the same.


quadkey/bq/test/kring_integration.js
```diff
@@ -18,18 +18,22 @@ describe('KRING integration tests', () => {
...
+        assert.deepEqual(rows[0].kring5,[12038154, 12038122, 12070890, 12103658, 12103690, 12103722, 12070954, 12038186, 12038154, 12038122, 12070890, 12103658, 12103690, 12103722, 12070954, 12038186, 12038154, 12038122, 12070890, 12103658, 12103690, 12103722, 12070954, 12038186, 12070922]);
+        assert.deepEqual(rows[0].kring6,[791032102930, 791032102898, 791040491506, 791048880114, 791048880146, 791048880178, 791040491570, 791032102962, 791032102930, 791032102898, 791040491506, 791048880114, 791048880146, 791048880178, 791040491570, 791032102962, 791032102930, 791032102898, 791040491506, 791048880114, 791048880146, 791048880178, 791040491570, 791032102962, 791032102930, 791032102898, 791040491506, 791048880114, 791048880146, 791048880178, 791040491570, 791032102962, 791032102930, 791032102898, 791040491506, 791048880114, 791048880146, 791048880178, 791040491570, 791032102962, 791032102930, 791032102898, 791040491506, 791048880114, 791048880146, 791048880178, 791040491570, 791032102962, 791040491538]);
```
Algunenano, line 36: I see lots of duplicates here, so something is going on.

quadkey/bq/sql/KRING.sql
```diff
@@ -15,15 +15,16 @@ AS """
...
-    let right     = sibling(quadint,'right').toString();
-    let topright  = sibling(right,'up').toString();
-    let downright = sibling(right,'down').toString();
-    let up        = sibling(quadint,'up').toString();
-    let down      = sibling(quadint,'down').toString();
-
-    return [left, topleft, downleft, right, topright, downright, up, down, quadint];
+    if(distance == null)
+    {
+        distance = 0;
```
Algunenano, line 20: Isn't 0 distance the own `quadint`? So IMO NULL -> 1, which should return the old behaviour.

Algunenano (changes requested):
> Notice that the test values where just reordered for the new implementation, but the values contained are the same. It's just that deepEqual complains if the order is not exactly the same.

Then sort them and check them sorted, otherwise the test is highly likely to break.

vdelacruzb:
- Updated the approach of the function and moved it to quadint.js.
- Added default distance as 1.
- Added tests.
- Fix integration_tests

Algunenano (approved):

PR #45, opened 2021-03-15
rename parameters

vdelacruzb:
The REFERENCE.md is being updated by @martejpad . She will include this little update


quadkey/bq/sql/LONGLAT_ASQUADINTLIST_RESOLUTION.sql
```diff
@@ -14,16 +14,16 @@ AS """
...
         return null;
     }
-    const zoom_min = parseInt(__zoom_min);
-    const zoom_max = parseInt(__zoom_max);
-    const zoom_step = parseInt(__zoom_step);
-    const resolution = parseInt(__resolution);
+    const zoomMin = parseInt(zoom_min);
+    const zoomMax = parseInt(zoom_max);
+    const zoomStep = parseInt(zoom_step);
+    const _resolution = parseInt(resolution);
```
Algunenano, line 20: What? So you removed it from the SQL parameter and then added it here which forces you to add it in 3 other places?

Algunenano (changes requested):

quadkey/bq/sql/LONGLAT_ASQUADINTLIST_RESOLUTION.sql
vdelacruzb: True, was mainly interested in removing "_" form the function notation rather than the inside and didn't came with a good name

vdelacruzb:
- Updated the function names for ST_BOUNDARY to be the same as we have in H3.
This PR goes together with:
https://github.com/CartoDB/carto-advanced-spatial-extension/pull/42

Algunenano:
> Updated the function names for ST_BOUNDARY to be the same as we have in H3.

We are going to need to start dropping renamed functions, as otherwise the old declarations will remain there. Can you take care of that, please?

vdelacruzb:
> > Updated the function names for ST_BOUNDARY to be the same as we have in H3.
> 
> We are going to need to start dropping renamed functions, as otherwise the old declarations will remain there. Can you take care of that, please?

As we have talked, just added a DROPS.sql where we can include the old function names.

Algunenano (approved):

PR #47, opened 2021-03-17
Fix st asquadint polyfill

vdelacruzb:
We were getting wrong values in quadint_polyfill, after investigating a bit further I saw that some int parsing were missing.

### DIFF
--- a/quadkey/bq/sql/KRING.sql
+++ b/quadkey/bq/sql/KRING.sql
@@ -19,6 +19,7 @@ if(quadint == null)
     {
         distance = 1;
     }
-    let neighbors = kring(quadint, distance);
+    const intDistance = parseInt(distance);
+    let neighbors = kring(quadint, intDistance);
     return neighbors.map(String);
 """;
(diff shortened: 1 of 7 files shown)


Algunenano (changes requested):
The fact that you are changing those number by these numbers gives me basically zero confidence. Can you add some tests that would have failed before and work now, please?

vdelacruzb:
I just added a test that would fail previously. The JS function wasn't returning the same values as the UDF, now it does.

Jesus89 (commented):
Looks good. But I have some questions:
- Should we review other functions (S2, H3, ...) to parseInt?
- In this case, should we also parse input params like `quadint`?

vdelacruzb:
> Looks good. But I have some questions:
> 
> * Should we review other functions (S2, H3, ...) to parseInt?
> * In this case, should we also parse input params like `quadint`?

* It would be good to check in another modules yes. I will add it to this review.
* The functions that accept quadint (bbox, parent, children, sibling...) call inside ZXYFromQuadint which already do it.

vdelacruzb:
- Applied changes to a S2 function.
Other functions in placekey and H3 where ok. I have followed the convention that we use in H3 and casting to Number instead of doing parseInt.

Jesus89 (approved):
LGTM

PR #55, opened 2021-04-05
fix st_boundary error

### LINKED ISSUE #54: ST_BOUNDARY generates error in BigQuery
Hi. I am using the bqcartoeu UDFs and have a table with H3 cell values generated from point data. I want to visualize different statistics on these cells, for example, number of observations per cell. I am trying to generate the geometry for the cells using:

`select bqcartoeu.h3.ST_BOUNDARY(bqcartoeu.h3.H3_FROMHEX(col_with_h3_hex_string))`

This gives an error:

`ST_GeogFromText failed: Invalid polygon loop: Edge 5 crosses edge 7 `

Unfortunately, I cannot localize which H3 values are causing the error, but I want to ask if the SAFE prefix could be added to the UDF where you are calling ST_GeogFromText.

I guess I am thinking it would make sense here?

https://github.com/CartoDB/carto-spatial-extension/blob/afb66d77dde3f1773a9bb877bd793cfcac30ae46/h3/bq/sql/ST_BOUNDARY.sql#L32

Thanks!

vdelacruzb:
Closes #54


Jesus89 (changes requested):
Could you add a test to cover this case? Thanks.

jgoizueta (approved):
LGTM

Jesus89:
Please, add the test and solve the conflicts

PR #57, opened 2021-04-06
Feature/ch144721/add h3 module to snowflake

vdelacruzb:
Notice that in Snowflake, H3 functions that returns array, do it in their hexadecimal form. This is due to finding lot of problems of parsing arrays, without later having subquery errors when calling these functions. We wrote yesterday to the Snowflake team to see if there is a good way to handle these cases.

Also as Snowflake functions support overload I allowed them to accept an BIGINT index and its hexadecimal form. 
*EDITED: the int conversions have been removed, this module works with hexadecimal now.


h3/sf/sql/VERSION.sql
```diff
@@ -0,0 +1,12 @@
...
+--
+-- Copyright (C) 2021 CARTO
+--
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE SECURE FUNCTION @@SF_DATABASEID@@.@@SF_SCHEMA_H3@@.VERSION()
+    RETURNS STRING
+    LANGUAGE JAVASCRIPT
+AS $$
+    return '3.7.0.0';
```
Jesus89, line 11: We should use a method here: h3Version()

Jesus89 (commented):
Code looks good.

Let's wait for the final changes in [this PR](https://github.com/CartoDB/carto-spatial-extension/pull/58) (to sync with this one) before merging.

h3/sf/sql/VERSION.sql
vdelacruzb: I have done it in one of my latest commits.
We were using directly the library "h3-js.umd.js" and now put together with our custom code, as we do in other modules.

h3/library/Makefile
```diff
@@ -2,14 +2,25 @@ TARGET = ../h3.js
...
+LINT_FILES = \
+	h3_version.js
 
 .PHONY: all
 all: ${TARGET}
-$(TARGET): $(AUXLIB)
-	$(CP) $(AUXLIB) $(TARGET)
+$(TARGET): $(JS_FILES)
+	rm -f $(TARGET)
+	touch $(TARGET)
```
Jesus89, line 20: Why touch after remove?

h3/sf/doc/REFERENCE.md
```diff
@@ -0,0 +1,374 @@
...
+
+**Example**
+
+```sql
+SELECT SFCARTO.H3.DISTANCE('847b591ffffffff', '847b59bffffffff');
+-- 1
+```
+
+{{% bannerNote type="note" title="tip"%}}
+If you want the distance in meters use [ST_DISTANCE](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_distance) between the cells ([ST_BOUNDARY](#st_boundary)) or their centroid.
```
Jesus89, line 293: We should point to snowflake docs here: https://docs.snowflake.com/en/sql-reference/functions/st_distance.html

h3/library/Makefile
vdelacruzb: No reason at all. I would say that was added to create back the file. But I see it's not needed. I've just removed it from all the modules.

h3/sf/doc/REFERENCE.md
vdelacruzb: Yes, thanks for the link.
We will also have to change "[Overview section](/spatial-extension-bq/spatial-indexes/overview/#h3)" at the beginning of the REFERENCE, in case links change when publishing snowflake doc.

Jesus89: I guess we will have another section in the future `spatial-extension-sf` with everything Snowflake.

Jesus89 (approved):
Great job!

PR #60, opened 2021-04-08
add functions to share

vdelacruzb:
I had still to update the shares when deploying.
Fix also a doc mistake on LONGLAT_ASH3 return type.


vdelacruzb:
I splitted the TOPARENT test and stopped checking the wide range of resolutions in Polyfill. It was too much for Snowflake since I was getting "JavaScript out of memory error: Query memory limit exceeded".

alex-graciano (approved):
Approved!

PR #61, opened 2021-04-08
Feature/ch144722/add s2 module to snowflake

vdelacruzb:
I also renamed some parameters, on LONGLAT_ASID to make them consistent with other modules, so will have to update doc.


jgoizueta (approved):
LGTM

PR #64, opened 2021-04-09
Release

Jesus89:
This PR updates the changelogs for the release and changes the format to https://keepachangelog.com/en/1.0.0/.


vdelacruzb:
It looks like the test fails, as we also test that the version is correct ^^'. So it has to normally be updated also there.

vdelacruzb (approved):
Nice! go for it.

PR #65, opened 2021-04-12
fix ci

vdelacruzb:


vdelacruzb:
After rebasing master and the release on friday, didn't update this test that makes the ci fail.

vdelacruzb:
Updated as well a bit the sh to match better how is done for bq.

alex-graciano (approved):
Looks good to me

PR #66, opened 2021-04-13
update function and tests

### LINKED ISSUE #63: Quadint prefix search
With quadkeys, it's easy to do a string prefix match to perform range queries (ie: `startswith(qk15, qk10)`). Is there a quick equivalent with quadints (ie: sql only)? Converting to quadkey or using `TOPARENT` (bit mask to extract level from parent id) would work, but rather avoid the js UDF in the join.

Use case:
- store a polygon as a collection of low zoom QK cells (perhaps varying zooms, but still non-overlapping) and join to points represented as a high zoom QK cell.

Would also be interested in this for s2 and h3 (though h3 non-nesting isn't ideal).

vdelacruzb:
I'm updating toparent to its sql version which is way faster.
Could have done more checks but this should be enough for covering the major issues without decreasing performance too much. The main difference with the previous is that this version accepts getting a quadint with the same resolution that the sent one and it would return the same value.

For instance:
-- bqcarto.quadkey.QUADINT_FROMZXY(3,1,1) -> 291
bqcarto.quadkey.TOPARENT(291, 3) -> 291

Closes #63 

The test will fail until the next PR is merged:
https://github.com/CartoDB/carto-spatial-extension/pull/65

### DIFF
--- a/quadkey/CHANGELOG.md
+++ b/quadkey/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+* TOPARENT based on entirely SQL implementation.
+
 ## [1.0.0] - 2021-03-31
 
 ### Added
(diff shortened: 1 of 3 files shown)


vdelacruzb:
This should be ready for review.

alex-graciano (approved):
Approved!

PR #59, opened 2021-04-06
First version of buffer in a new module for transformation

jatorre:
A new module called transformation with buffer as its first function

Issue #46, opened 2021-03-16
Function not found: jslibs.turf.ST_BUFFER 


cvargas-geo:
I am executing a query in BigQuery, but this is not working, I get the following error:

`Function not found: jslibs.turf.ST_BUFFER at [2: 3]`

But this line works perfectly:
`SELECT jslibs.turf.ST_BUFFER (ST_GEOGPOINT (-33.2187979, -70.6843369), 1000, 'meters', 10) AS geo;`

![image](https://user-images.githubusercontent.com/78808547/111356678-5cb43700-8667-11eb-870c-c4f77ecd4fef.png)

![image](https://user-images.githubusercontent.com/78808547/111357271-eebc3f80-8667-11eb-95b8-b78955daea7d.png)

Algunenano:
Hi there, Is it possible that you are not running your query in the US region?

Functions can't be shared across regions so the extension currently only support the US (`jslibs.turf.ST_BUFFER`) and EU (`jslibs.eu_turf.ST_BUFFER`) multi regions. To use it on other regions or under VPC you'll need to deploy them yourself in your own project.

vdelacruzb:
Closing this issues as we finally have deployed ST_BUFFER under the new repo structure.
Please use:
`bqcarto.transformation.ST_BUFFER`
and in Europe:
`bqcartoeu.transformation.ST_BUFFER`

Example:
`SELECT bqcarto.transformation.ST_BUFFER(ST_GEOGPOINT(-74.00, 40.7128), 1, 'kilometers', 10);`

PR #68, opened 2021-04-16
Feature/ch147306/add turf st buffer

vdelacruzb:
Just saw a post calling this module "TRANSFORMATION".  Should I change it?
- The js test for turf that i'm including are directly taken from turf, so if we update the version we should update them as well.


vdelacruzb:
Closes #46
Closes #59

Jesus89 (approved):
LGTM.

Could you add the reference for the version function?

vdelacruzb:
I just added the version to the reference and merging.

PR #69, opened 2021-04-16
Release

Jesus89:


vdelacruzb (approved):
LGTM.
Let's see if those test pass with the changes on versions and let's go for it.

transformation/library/package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "transformation_linter",
+  "name": "transform_linter",
```
vdelacruzb, line 2: this could be still transformation_linter. 
But it's a quick change for next times

PR #71, opened 2021-04-21
Improve CI

Jesus89:
- Parallelize tests: linter, unit, integration-bq, integration-sf
- Check diff in GH Actions
- Check diff modules in Makefile
- Minor fixes Makefiles and env script

NOTE: regarding git diff, it would be nice to detect it earlier so the job can be canceled (if this is possible in GH Actions). I have added a variable `ALL` in the Makefile to force check all the modules. Should we run this on CI in `master` branch?


transformation/Makefile
```diff
@@ -1,6 +1,5 @@
-DBS = "bq/"
 LIBRARY = "library/"
-ALL_SUBFOLDERS = $(DBS) $(LIBRARY)
+ALL_SUBFOLDERS = $(LIBRARY) "bq/" "sf/"
```
vdelacruzb, line 2: This has no "sf" for now. Don't know if it will give problems when cleaning. Nice that you noticed that we were missing it in the others.

vdelacruzb (approved):
LGTM.
I would remove the "sf" in transformation just in case. And I think it's good to run ALL in the master branch just to be super sure that we are stable.

transformation/Makefile
Jesus89: No problem with this

Jesus89:
I have improved the flow to check the "testability" first, then the linter, and then unit and integration tests (bq, sf):



I have added also a diff checker (https://github.com/marketplace/actions/get-diff-action). It checks changes in Makefiles, JS, or SQL files and the main Makefile detects the modules that have changed. Then, CI tests are faster, cheaper, and more efficient :smile:

PR #73, opened 2021-04-23
Add setool

Jesus89:
Custom generator tool as an alternative to "copy and paste the skel directory and do all the replacements manually", so contributing becomes much easier and faster.

NOTE: check `hygen` in the future.


alex-graciano (approved):
LGTM

tools/setool/lib/create-module/generator.js
```diff
@@ -0,0 +1,922 @@
...
+--
+-----------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION \`@@BQ_PROJECTID@@.@@BQ_DATASET_${uname}@@.VERSION\`()
+    RETURNS STRING
+    DETERMINISTIC
+    LANGUAGE js${islib(`
+    OPTIONS (library=["@@${uname}_BQ_LIBRARY@@"])`)}
+AS """
+    return ${library ? `${lname}Version()` : `'1.0.0'`};
```
jgoizueta, line 466: This looks good to me, but the code generated for the `constructor` module didn't have this right, have you already fixed it?

Jesus89: I think the code in the `constructors` module is pure SQL, right?

alex-graciano: The code developed so far is pure SQL, however for the turf functions I'm pretty sure we are requiring JS code.

PR #76, opened 2021-04-28
Added ST_TILEENVELOPE to "constructors" module

alex-graciano:
The original PostGIS function (https://postgis.net/docs/ST_TileEnvelope.html) also includes an extra `margin` parameter to grow a tile by the given percentage. As we stated in Slack, the current implementation uses just the (x,y,z) index to extract the polygon but in case the `margin`parameter is further required, this implementation is not valid and we must go with a previous step for extracting the bounding box. We can use something like this:

```sql
    SELECT `@@BQ_PROJECTID@@.@@BQ_DATASET_CONSTRUCTORS@@.ST_MAKEENVELOPE`(
        bounds[OFFSET(0)],
        bounds[OFFSET(1)],
        bounds[OFFSET(2)],
        bounds[OFFSET(3)]
    )
    FROM (
        SELECT `@@BQ_PROJECTID@@.@@BQ_DATASET_QUADKEY@@.BBOX`(
            `@@BQ_PROJECTID@@.@@BQ_DATASET_QUADKEY@@.QUADINT_FROMZXY`(zoomLevel,xTile,yTile)) AS bounds
    )
```

The problem is the `ST_MAKEPOLYGON` used in our `ST_MAKEENVELOPE` has some issues with dealing with lower zoom levels, splitting the resulting polygon in a weird fashion, e.g. giving this result for the (0,0,0) tile:

`GEOMETRYCOLLECTION(POLYGON((0 0, 120 0, -120 0, 0 0)), POLYGON((0 0, -120 0, 120 0, 0 0)))`

In conclusion, I would keep the signature as it is, just tanking the tile index.

### DIFF
--- a/constructors/CHANGELOG.md
+++ b/constructors/CHANGELOG.md
@@ -9,3 +9,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * Initial implementation of the module.
 * Add ST_MAKEENVELOPE function
+* Add ST_TILEENVELOPE function
--- a/constructors/bq/Makefile
+++ b/constructors/bq/Makefile
@@ -27,6 +27,9 @@ endif
 ifndef BQ_BUCKET_PUBLIC
 	$(error BQ_BUCKET_PUBLIC is undefined)
 endif
+ifndef BQ_DATASET_QUADKEY
+	$(error BQ_DATASET_QUADKEY is undefined)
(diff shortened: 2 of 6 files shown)


Jesus89 (approved):
LGTM

Jesus89:
Tested in staging :+1:

vdelacruzb (approved):
Nice

PR #78, opened 2021-04-28
Staging custom deploy

ilbambino:


Jesus89 (approved):

Jesus89:
Hey @ilbambino it looks like the SA has not enough permissions to create the project:

```
ERROR: (gcloud.beta.billing.projects.link) User [spatial-extension-ci@bqcartoci.iam.gserviceaccount.com] does not have permission to access project [bqcartost-core-78] (or it may not exist): The caller does not have permission
```

Jesus89 (approved):

Jesus89 (approved):

Jesus89:
The creation of the project is an async process, so we need to make sure the project is created (with a loop or something similar).

There is also another issue regarding the bucket. Not sure if it's not created or it requires an extra step.

```NotFoundException: 404 The destination bucket gs://carto-ext-ci-public-bqcartost-core-78 does not exist or the write to the destination must be restarted```

Jesus89 (approved):

Jesus89 (approved):

Jesus89 (approved):

Jesus89 (approved):

Jesus89:
There was a problem with the SA permissions that now is fixed.

NOTE: the failing Staging / deploy-bq is because of a project without bucket. But we have validated that it works in previous tests in this PR.

PR #79, opened 2021-04-28
Feature/ch149081/add turf functions

vdelacruzb:
Find here the public part of the requested turf functions.
I added the same tests that turf adds to the js part so we check that the library is well constructed.
Also I'm concerned about not missing anything given the fact that some CI and makefiles changed while I was working on the functions.
Some of the functions include "Optional" parameters what means, they are mandatory ^^' but if users set the value to NULL the value is set to default.


vdelacruzb:
As discussed in the meeting. I'm going to raise transformations and constructors version to 1.1.0 instead of 1.0.1

constructors/bq/test/package-lock.json
```diff
@@ -1,1036 +1,1117 @@
 {
-    "name": "constructors_bq_integration_tests",
```
alex-graciano, line 2: IMO I think this file can be added to `.gitignore`

constructors/library/test/bezierspline_test/out/bezierIn.json
```diff
@@ -0,0 +1,541 @@
+{
```
alex-graciano, line 1: Maybe the test output can add noise to the repo. Are they being used in CI or somewhere?

measurements/CHANGELOG.md
```diff
@@ -4,8 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.0.0] - 2021-04-16
+## [1.0.0] - 2021-04-29
```
alex-graciano, line 7: Are we considering the new additions to be in the same version? It is Ok to me.

constructors/library/test/bezierspline_test/out/bezierIn.json
vdelacruzb: yes, these tests contain fixtures and they normally pass the in/ to the function and compare that the out/ matches what is expected

constructors/bq/test/package-lock.json
vdelacruzb: yes, I just added it and seems to pass all the tests.

measurements/bq/doc/REFERENCE.md
```diff
@@ -0,0 +1,99 @@
...
+
+{{% bannerNote type="code" %}}
+measurements.ST_ANGLE(startPoint, midPoint, endPoint, mercator)
+{{%/ bannerNote %}}
+
+**Description**
+
+Finds the angle formed by two adjacent segments defined by 3 points. The result will be the (positive clockwise) angle with origin on the startPoint-midPoint segment, or its explementary angle if required. https://github.com/Turfjs/turf/tree/master/packages/turf-angle
+
+* `startPoint`: `GEOGRAPHY` Start Point Coordinates.
```
alex-graciano, line 15: We, including myself, should be consistent with starting the parameter description with upper or lower case. It should be consistent at least in the same `REFERENCE` file. For instance:

### ST_ANGLE
* `startPoint`: `GEOGRAPHY` Start Point Coordinates.

### ST_AZIMUTH
* `startPoint`: `GEOGRAPHY` starting Point.

measurements/CHANGELOG.md
alex-graciano: Clarified with @vdelacruzb. It is actually a new module.

transformations/Makefile
```diff
@@ -0,0 +1,26 @@
+LIBRARY = "library/"
+ALL_SUBFOLDERS = $(LIBRARY) "bq/" "sf/"
```
alex-graciano, line 2: The reference to sf/ directory is generated by setool. Should we remove it from the generator in case just BQ project is selected @Jesus89?

transformations/bq/sql/_DROPS.sql
```diff
@@ -0,0 +1,7 @@
+-----------------------------------------------------------------------
+--
+-- Copyright (C) 2021 CARTO
+--
+-----------------------------------------------------------------------
+
+DROP SCHEMA IF EXISTS `@@BQ_PROJECTID@@.transformation`;
```
alex-graciano, line 7: `@@BQ_PROJECTID@@.transformations`? Or simply `@@BQ_PROJECTID@@.@@BQ_DATASET_TRANSFORMATIONS@@`

alex-graciano (commented):
Good job! Just take a look to the points to be reviewed and for me it is ready to merge.

transformations/bq/sql/_DROPS.sql
vdelacruzb: I set it has @@BQ_PROJECTID@@.transformation as we previously had this module called `transformation`, and it's deployed in both staging and production. As we want to drop the previous one that's why I don't had the `s` or `@@BQ_DATASET_TRANSFORMATIONS@@`

alex-graciano: Ok! I imagined something like that.

transformations/Makefile
Jesus89: We adjust this, but now it's not a problem because there is a check (if [ -d $${s} ];)

Jesus89 (approved):
Outstanding job! LGTM

Notes:
- I'm not sure about passing `NULL` to provide the default value. We need to check if there's a standard way of doing this.
- We should revisit also the Unit tests fixtures (use mocks), package-lock.json, and some details from the `skel` template. But let's move forward with this and review this in future iterations.

constructors/bq/doc/REFERENCE.md
```diff
@@ -4,6 +4,46 @@
 
 This module contains functions that create geographies from coordinates or already existing geographies.
 
+### ST_BEZIERSPLINE
+
+{{% bannerNote type="code" %}}
+constructors.ST_BEZIERSPLINE(geog, sharpness)
```
Jesus89, line 10: Missing `resolution`

vdelacruzb: True, I added this parameters a bit after and missed to update the REFERENCE. Now should be good

Jesus89:
Tested in staging :+1:

PR #95, opened 2021-05-12
Code structure improvements accessors

vdelacruzb:


Jesus89 (commented):
LGTM.

Note: I would move the clean to the Makefile instead of the package.json files

PR #96, opened 2021-05-13
Code structure improvements h3

vdelacruzb:


vdelacruzb:
@Jesus89 I added the code on top of the base branch but seems that constructors fails because of not having quadkey. Let's review this once quadkey is merged.

PR #117, opened 2021-05-27
Optimize H3 bundle size

vdelacruzb:
- Fixed hexring having wrong argument.
- The external lib where modified and created manually from h3-js (https://github.com/uber/h3-js/). It could be automatized until certain point, but it would require installing docker, and lot of configs. And there is some modifications that have to be manual if we really want to optimize at maximum the dependencies that we take. As we are not directly using npm, should I add the LICENSE?
- I was thinking in doing an external SQL calls to h3.ISVALID, to not add it has a dependency and save some space. But tested that adding it to the bundle doesn't really increase the size, so I keep it.
- Having another pattern `externalLib` to make eslint ignore libs that are taken from outside, or that are already minified. I accept suggestions if the naming could be other.
- Would like to improve having capital letters as the way of naming the library in the functions. Example, @@SF_LIBRARY_compact@@ -> @@SF_LIBRARY_COMPACT@@. You will notice that in this case we don't have a general pattern @@SF_LIBRARY_CONTENT@@ as we need to know what library to paste. We could have it if by default we created a library per function and would need some directories restructuring, each function in a different folder and containing doc, sql, lib and test, but it's not our case right now.
- Have kept the possibility of having a lib folder to not break other modules. So the makefile check if the lib folder exists to look for index.js and if libs folder exist to check all the libraries.


Jesus89:
Wow, great job!

- I have created a ticket to review the LICENSE files in every module, so don't worry about it in this PR.
- I can push a commit with the uppercase change for you

I'm going to think a bit about removing duplicated code, but overall looks very good

Jesus89:
I have updated the implementation to reduce duplicated code in a PR: https://github.com/CartoDB/carto-spatial-extension/pull/122
- Having a common h3core customized to bind only exported functions dynamically.
- Import bindings from the original source code.

The build system includes now PRE_BUILD and POST_BUILD scripts that h3/snowflake uses to copy and remove the h3core_custom file to every lib directory. I have updated the name of the files to use "custom" instead of "externalLib".

BEFORE:
- index: 142.51 kB

AFTER:
- ash3_polyfill: 95.99 kB
- boundary: 55.36 kB
- compact: 41.24 kB
- distance: 50.69 kB
- hexring: 50.64 kB
- ispentagon: 33.72 kB
- isvalid: 34.03 kB
- kring: 53.28 kB
- longlat_ash3: 51.16 kB
- tochildren: 36.37 kB
- toparent: 34.95 kB
- uncompact: 37.09 kB

Jesus89 (approved):
LGTM

PR #123, opened 2021-06-14
fix shares issues

vdelacruzb:
- Fix some private functions eligible for sharing. 
- Fix one public function not eligible for sharing.


alex-graciano (approved):
LGTM!

PR #125, opened 2021-06-15
Support new production projects

Jesus89:
Add new projects for price plans:
- [x] Update CD scripts
- [x] Update permissions scripts
- [x] Change test-integration target names
- [x] Update doc examples to use the `customSelector` component


README.md
```diff
@@ -12,7 +12,7 @@ SELECT jslibs.h3.ST_H3(ST_GEOGPOINT(10,10),11)
 it is now:
 
 ```sql
-SELECT bqcarto.h3.ST_ASH3(ST_GEOGPOINT(10,10),11) 
+SELECT carto-os.h3.ST_ASH3(ST_GEOGPOINT(10,10),11)
```
vdelacruzb, line 15: I know that we don't have much choice on this. But do you think these will be the last naming in the project? We already received some complains in an issue about users that didn't move still from jslibs. So I'm worried users can fill that they have to maintain their projects constantly in order to use our functions.

vdelacruzb (approved):
Great job on this! Let's see how deployment goes.

README.md
Jesus89: Yep. The point is that now we're moving to price-plan projects so it looks like these will be the final ones. The product team needs to define a transition to the new projects for the current users and this must be aligned with the cloud-native strategy. When a user changes the price, plan he/she will need to use the corresponding project. This is how it's defined by the product team.

We will maintain the `bqcarto` and `bqcartoeu` ones with updates, but eventually, at some point, they will be deprecated them. cc @martejpad

PR #126, opened 2021-07-16
First version of the package installation system (Core project)

alex-graciano:


common/bigquery/Makefile
```diff
@@ -26,7 +26,7 @@ REPLACEMENTS = -e 's!@@BQ_PREFIX@@!$(BQ_PREFIX)!g' -e 's!@@BQ_LIBRARY_BUCKET@@!$
 .PHONY: help lint lint-fix build test-unit test-integration test-integration-full deploy clean clean-deploy
 
 help:
-	echo "Please choose one of the following targets: lint, lint-fix, build, test-unit, test-integration, test-integration-full, deploy, clean, clean-deploy"
+	echo "Please choose one of the following targets: lint, lint-fix, build, test-unit, test-integration, test-integration-dry, deploy, clean, clean-deploy, serialize_package"
```
jgoizueta, line 29: Hey you've seemed to revert `test-integration-full` to `-dry`

Makefile
```diff
@@ -5,13 +5,24 @@ export GIT_DIFF ?= off
...
-lint lint-fix build test-unit test-integration deploy clean clean-deploy:
 	if [ "$(CLOUD)" = "bigquery" ] || [ "$(CLOUD)" = "snowflake" ]; then \
+		if [ "$@" = "serialize_package" ]; then \
+			rm -rf dist/; \
+			mkdir dist/; \
+		fi; \
 		for module in `node scripts/modulesort.js`; do \
 			echo "> Module $${module}/$(CLOUD)"; \
 			$(MAKE) -C modules/$${module}/$(CLOUD) $@ || exit 1; \
+			if [ "$@" = "serialize_package" ]; then \
```
jgoizueta, line 20: Why are we doing all this in the root Makefile? Couldn't we handle it in the serialize_package task in common/bigquery?

If there's a good reason to put it here I would consider having a separate target for serialize_package here even if it duplicates the `for module...` to keep the other targets simple.

jgoizueta (approved):
LGTM
I left a note to consider refactoring the Makefile part here and in the core ext, but I consider it optional.

Makefile
```diff
@@ -5,13 +5,24 @@ export GIT_DIFF ?= off
 .PHONY: help lint lint-fix build test-unit test-integration test-integration-full deploy clean clean-deploy
 
 help:
-	echo "Please choose one of the following targets: lint, lint-fix, build, test-unit, test-integration, test-integration-full, deploy, clean, clean-deploy"
+	echo "Please choose one of the following targets: lint, lint-fix, build, test-unit, test-integration, test-integration-dry, deploy, clean, clean-deploy, serialize_package"
```
Jesus89, line 8: [Here the same](https://github.com/CartoDB/carto-spatial-extension/pull/126/files#r672871164). We should also follow the naming convention here: `serialize-package`, but I think `build-package` should be more convenient

Jesus89: Having the advanced PR context I think `serialize-package` is correct

Makefile
alex-graciano: I put this target in the root Makefile in order to avoid code duplication about the module iteration. In this way, the schema of this target is in line with the already introduced (the target in common/bigquery manages just one module). However, as you said maybe an intermediate target could be sensible. Working on it.

Jesus89 (changes requested):
A few missing changes:

- Rename `serialize_package` to `serialize-package`
- Change `test-integration-dry` to `test-integration-full`
- Generate temporary `bundle_pkg.sql` into module `dist/`.
- Remove line `echo "" >> bundle_pkg.sql; \`, is this really needed?
- Remove "snowflake" cloud in `serialize-package` rule (it is only available for bigquery)
- Consider moving main `serialize-package` commands to a script in `common/bigquery` to clean the Makefile

Jesus89:
I did some minor naming changes

Jesus89 (approved):
Great job. LGTM.

I'll merge this one and the update and merge the advanced one

PR #129, opened 2021-07-23
Add Redshift tooling/structure

Jesus89:
Add the structure, tools, and scripts for Redshift modules. Using quadkey module to test the VERSION function.


vdelacruzb:
The added options in the flake8 config file `tox.ini`are there to make it compliant with  `black`. Notice that black doesn't like [single quotes](https://github.com/psf/black/issues/594). Should we stick to double quotes? the other option would be to add the argument `skip-string-normalization` to the black command but we would take full control on the strings formatting

vdelacruzb (approved):
Deploying to staging

Jesus89:
Nice change using the `aws --query` arg. I'll do a final review and merge

common/redshift/Makefile
```diff
@@ -87,10 +88,9 @@ storage-upload: build
 		$(AWS) s3 cp $$f $(AWS_S3_BUCKET) || exit 1; \
 	done
 	for f in $(notdir $(wildcard dist/*.zip)); do \
-		STATEMENT_SQL="CREATE OR REPLACE LIBRARY $(RS_SCHEMA) LANGUAGE plpythonu FROM '$(AWS_S3_BUCKET)$(notdir $$f)' CREDENTIALS 'aws_access_key_id=$(AWS_ACCESS_KEY_ID);aws_secret_access_key=$(AWS_SECRET_ACCESS_KEY)';"; \
+		STATEMENT_SQL="CREATE OR cREPLACE LIBRARY $(RS_SCHEMA) LANGUAGE plpythonu FROM '$(AWS_S3_BUCKET)$(notdir $$f)' CREDENTIALS 'aws_access_key_id=$(AWS_ACCESS_KEY_ID);aws_secret_access_key=$(AWS_SECRET_ACCESS_KEY)';"; \
```
vdelacruzb: little typo `cREPLACE`

Jesus89:
OK. So after reviewing the AWS output I think this is ready to be merged: not it outputs the query and the error if exists, and also stops the execution if there's an error.

PR #135, opened 2021-07-29
Allow sharing with multiple shares in Snowflake

vdelacruzb:
- We were losing links between functions and shares on each deployment. We are using the `ADVANCED_SPATIAL_EXTENSION` share in the marketplace and each time we deploy core very likely the links to that share were broken. Now each time we deploy core it will also add the core functions to the `ADVANCED_SPATIAL_EXTENSION` share so the link shouldn't be lost anymore. Would be awesome as well to request the SF team for a `COPY GRANTS` option.

- I use this PR also to add `USE ROLE ACCOUNTADMIN` in the `share-remove` since it should be crashing.


Jesus89 (approved):
Code LGTM.

There are a couple of issues:
- Conflicts must be solved
- You need to force push this branch with your user in order to pass CLA

PR #140, opened 2021-07-30
Fix Quadint ST_BOUNDARY

vdelacruzb:
- Fix issues with __BBOX components returning wrong values.
- Format a bit the functions.
- Reorder the values of BBOX to [xmin, ymin, xmax, ymax]


vdelacruzb:
- Updated integration test. I didn't revert to the previous version since as we are using `ST_MAKEENVELOPE` now, the values are shifted with the previous version, but they are the same.
Also regarding the BBOX test as we are performing the operations through SQL and not JS the las decimals numbers is what changes.

jgoizueta (commented):
Nice! and you fixed the __BBOX__N, great!

vdelacruzb:
Closing this PR and reopening on https://github.com/CartoDB/carto-spatial-extension/pull/141

PR #150, opened 2021-08-10
Support geometry collection in Quadint polyfill

vdelacruzb:


modules/quadkey/bigquery/sql/ST_ASQUADINT_POLYFILL.sql
```diff
@@ -17,6 +17,22 @@ AS """
...
+CREATE OR REPLACE FUNCTION `@@BQ_PREFIX@@quadkey.__geometrycollection`
+(geojson STRING)
+RETURNS STRING
+DETERMINISTIC
+LANGUAGE js
+OPTIONS (library=["@@BQ_LIBRARY_BUCKET@@"])
+AS """
+    const pol = JSON.parse(geojson);
+    return JSON.stringify(quadkeyLib.geometrycollection(pol))
+    // Iterate through collection features
```
Jesus89, line 29: Why are all the comments here?

vdelacruzb: It was early WIP, just in case Alex had to continue. The function works now, just have to add integration tests.

By the way @alex-graciano I have seen that we have the same problem with H3 polyfill. Do you want to tackle that one? It's not crashing but it returns an empty array.

vdelacruzb:
Made tests in SF use `ST_ASQUADINT_POLYFILL` and also added duplicated features to tests to double check that the resulting array doesn't change as we remove duplicates.

Jesus89 (approved):
LGTM!

Jesus89:
I've tested this in staging and found the following error:

```sql
SELECT quadkey.ST_ASQUADINT_POLYFILL(geom,16) FROM `cartobq.testtables.polygons_800k` LIMIT 1000
-- Error: NULL argument passed to UDF at UDF$1(STRING, INT64) line 3, columns 8-9
```

This is because the dataset contains `GEOMETRYCOLLECTION EMPTY` and this ends up in NULL GeoJSON. I think we should cover this case to avoid the bug, and return `[]` in that case. cc @martejpad

vdelacruzb:
> I've tested this in staging and found the following error:
> 
> ```sql
> SELECT quadkey.ST_ASQUADINT_POLYFILL(geom,16) FROM `cartobq.testtables.polygons_800k` LIMIT 1000
> -- Error: NULL argument passed to UDF at UDF$1(STRING, INT64) line 3, columns 8-9
> ```
> 
> This is because the dataset contains `NULL` geographies and this ends up in `NULL` GeoJSON. I think we should cover this case to avoid the bug, and return `[]` in that case. cc @martejpad

Definitely I remember that in some cases we are returning errors in Quadint but for this one, it should be an empty array. Updating it.

vdelacruzb:
The NULL issue is related with https://github.com/CartoDB/carto-spatial-extension/issues/148.

I thought Bigquery's [Safe Prefix](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#safe_prefix) was also working with UDFs. Should we request that? If it was supported all the functions should return Error by default.

Jesus89:
OK then. We can merge it as it is and create a ticket in CH to update this behavior in all the functions if required and possible.

Jesus89:
But remember to update the changelog first

PR #155, opened 2021-08-13
Support geometry collection in H3 polyfill

vdelacruzb:


vdelacruzb:
No performance changes have been applied to this PR maybe in the future we can try to improve it by using a more imperative programming.

Jesus89 (approved):
LGTM!

I would clarify a bit the notation used in the tests.

Jesus89:
Please, update the changelog and the version number too

vdelacruzb:
- Added a Multipolygon to the GEOMETRYCOLLECTION test.
- Updated the comments as they could be confused.
- Removed the comment on `-- 5 is a...` as the GEOMETRYCOLLECTION is not tata composed of other test geoms anymore.

PR #161, opened 2021-09-09
Add quadkey module to redshift

vdelacruzb:
- Add quadkey functions:
  - The functions returning arrays KRING, TOCHILDREN, ST_ASQUADINT_POLYFILL are returning SUPER, that should work like them.
  - The functions returning Geometry ST_BOUNDARY are returning VARCHAR for now. If Redshift doesn't create a `ST_GEOMFROMGEOJSON`, we will convert geojsons to wkb/wkt.
  - As in SQL functions we don't have named parameters I'm leaving a comment just below, maybe we can use that style or similar.
  - The polyfill function supports Geometry Collections as in other warehouses. The tilecover code as been ported from https://github.com/mapbox/tile-cover and https://github.com/mapbox/tilebelt.
  - Mercantile already contained functions for parent, children, neighbors but used the same that we created in other warehouses for convinience.
 
- Modified the Makefile:
  - So we can create structure of folders inside lib, and not insert everything in the same __init__.py
  - Added -rP to pytest in case we want to debug the test and see the prints.
  - If found venv2 installs `python2_module_requirements.txt` which are the module specific dependencies, used for building the library and unit tests.
  - If found venv3 installs `python3_module_requirements.txt` which are the module specific dependencies used only for integration tests.
  - Library installer runs through venv3 and install the modules in `python2_module_requirements`.

[ch176381]


vdelacruzb:
BTW when we merge this PR, we should also take a look to advanced to update the MAKEFILE, add library_installer, etc.

modules/quadkey/redshift/sql/LONGLAT_ASQUADINT.sql
```diff
@@ -0,0 +1,17 @@
+----------------------------
+-- Copyright (C) 2021 CARTO
+----------------------------
+
+CREATE OR REPLACE FUNCTION @@RS_PREFIX@@quadkey.LONGLAT_ASQUADINT
+(longitude FLOAT8, latitude FLOAT8, resolution INT)
+-- (longitude, latitude, resolution)
```
Jesus89, line 7: Comment not needed

modules/quadkey/redshift/sql/ST_ASQUADINT_POLYFILL.sql
```diff
@@ -0,0 +1,43 @@
+----------------------------
```
Jesus89, line 1: Remove one copyright header

Jesus89 (commented):
Outstanding work!

I have left some comments to review. Tomorrow I'll test it in staging

Jesus89:
The tests fail because `storage-upload` requires `venv3`. I'll upload the fix

Jesus89 (approved):
LGTM!

Jesus89:
Re-testing staging deployment

Jesus89 (approved):

PR #165, opened 2021-09-11
s2 in Redshift

arredond:
First working functions using `s2sphere` Python library:

- LONGLAT_AS_ID: Returns the integer ID of a cell at a certain longitude, latitude and resolution level
- GET_RESOLUTION: Returns the resolution (level) of a cell, given an integer ID
- TOPARENT: Returns the parent ID of a cell, optionally at a certain resolution level
- TOCHILDREN: Returns a comma-separated string with all child IDs, optionally at a certain resolution level
- ST_BOUNDARY: Returns the boundary of a cell, given an integer ID. This is still WIP, see the below comments for doubts on what format to use to return the boundary

To do:
- [x] More functions: region coverer (polyfill), etc
- [x] Unit tests
- [x] Integration tests
- [x] Docs


arredond:
Some doubts I'm having:

- **Geometrical formats**: Since types are limited in RS, I suppose the best way to go is to output geometries as text/bytes in either GeoJSON, WKT, WKB or plain concatenated strings (such as bboxes). How should we proceed? I see in the Quadkey module we've gone with GeoJSON + bbox strings but these have some gotchas (see next point)

- S2 cells are based on a spherical geometry where edges connecting vertices are great circles (shortest paths on a sphere, not a plane). Therefore, when returning a cell's boundary, we have to choose whether to return just the vertices (and let the client join them with great circles) or returning the whole thing projected on a plane (remember the GeoJSON spec is only for WGS84 lat/long proj)

vdelacruzb:
> Some doubts I'm having:
> 
> * **Geometrical formats**: Since types are limited in RS, I suppose the best way to go is to output geometries as text/bytes in either GeoJSON, WKT, WKB or plain concatenated strings (such as bboxes). How should we proceed? I see in the Quadkey module we've gone with GeoJSON + bbox strings but these have some gotchas (see next point)

Actually Redshift have a GEOMETRY type. In the case of quadkey I decided to go for a VARCHAR for the moment waiting for them to answer if they have something like `ST_GEOMFROMGEOJSON` in mind. They already have `ST_GeomFromText`for instance. 

> * S2 cells are based on a spherical geometry where edges connecting vertices are great circles (shortest paths on a sphere, not a plane). Therefore, when returning a cell's boundary, we have to choose whether to return just the vertices (and let the client join them with great circles) or returning the whole thing projected on a plane (remember the GeoJSON spec is only for WGS84 lat/long proj)

On this I would follow what we do in other modules. I believe we return a geojson with just the vertices. For instance in SF: 

`select sfcarto.s2.st_boundary(sfcarto.s2.LONGLAT_ASID(100, 30, 5));`

Returns:
```
{
  "coordinates": [
    [
      [
        97.78514014329619,
        28.450360919447153
      ],
      [
        97.78514014329619,
        31.391909433476105
      ],
      [
        100.61965527615513,
        31.188056199371985
      ],
      [
        100.61965527615513,
        28.25839533868567
      ],
      [
        97.78514014329619,
        28.450360919447153
      ]
    ]
  ],
  "type": "Polygon"
}
```

> * S2 defines cell IDs in two ways: a 64-bit integer ID and a hex-encoded token of that same integer ID. Do we have any preferences? The BQ S2 module just uses the integer ID but we can have some methods to convert between both since the token is also widely used (for example, [here](https://s2.sidewalklabs.com/regioncoverer/?center=40.416881%2C-3.703462&zoom=20&cells=0d42287e177))

I'm not totally sure if at some point we also had the hex version an we removed. Was looking at jslibs but maybe it wasn't there still https://github.com/CartoDB/carto-spatial-extension/tree/18f96426ee864a8ae76488ca16a297b0b712ecdc/sql/s2. I guess for now we prefer the integer version but we can maybe raise this topic in slack to see if we had the hex conversion to all warehouses.

arredond:
After talking about this, we'll be returning a plain WKT with the vertices and expect the client to correctly render the cell boundaries with geodesic edges.

We'll also have to add a note about this in the docs

common/redshift/Makefile
```diff
@@ -142,6 +142,7 @@ venv3:
 	. $(VENV3_ACTIVATE) && \
 	python -m pip install -U pip -q && \
 	pip install -r $(COMMON_DIR)/redshift/python3_requirements.txt -q && \
+	([ ! -f python3_requirements.txt ] || pip install -r python3_requirements.txt -q 2>/dev/null) && \
```
Jesus89, line 145: This line is not needed.

modules/s2/redshift/python3_requirements.txt
```diff
@@ -0,0 +1 @@
+s2sphere==0.2.5
```
Jesus89, line 1: This is not needed.

modules/s2/redshift/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy==1.16.6
```
Jesus89, line 1: sNumpy is not needed here because it's installed by default in the Redshift clusters. Remember to update the shasum file after this change.

arredond: Yep, I removed the dependency from the lib too so thanks for this one

Jesus89 (commented):
**OK functions**
- LONGLAT_ASID
- ST_BOUNDARY
- TOKEN_FROMID
- UINT64REPR_FROM_ID -> UINT64REPR_FROMID
- VERSION

**Missing functions**
- HILBERTQUADKEY_FROMID
- ID_FROMHILBERTQUADKEY
- ID_FROMTOKEN
- ID_FROMUINT64REPR
- ST_ASID

**Extra functions**
- TOPARENT
- TOCHILDREN
- GET_RESOLUTION -> RESOLUTION
- ST_POLYFILL_BBOX

common/redshift/Makefile
arredond: Let me explain the logic behind this one: s2sphere is imported at the top of the s2Lib `__init__.py`. The `version` integration test imports the `__version__` from `__init__.py` so, when running the integration tests, `s2sphere` has to be installed in the `venv3`

arredond: I suppose this is why @vdelacruzb put all the imports inside functions in the quadkey RS module

modules/s2/redshift/python3_requirements.txt
Jesus89: Your idea of using _version.py file with __version__ is the best option :+1:

modules/s2/redshift/lib/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '1.0.0'
```
Jesus89, line 1: Use `_version.py` which is more standard

modules/s2/redshift/test/integration/test_VERSION.py
```diff
@@ -0,0 +1,10 @@
+from test_utils import run_query
+
+# Initialize variable just to avoid linter complaints
+__version__ = None
+exec(open('./lib/version.py').read())
```
Jesus89, line 5: Could you use this here?

```py
from lib._version import __version__
```

modules/s2/redshift/test/unit/test_version.py
```diff
@@ -0,0 +1,12 @@
+from lib import s2Lib, __version__
```
Jesus89, line 1: Here you can import directly from the _version file

```py
from lib import s2Lib
from lib._version import __version__
```

modules/s2/redshift/test/integration/test_VERSION.py
arredond: I don't think so, this still tries to run `__init__.py`.

Took this from this answer: https://stackoverflow.com/a/16084844/8159847

modules/s2/redshift/lib/version.py
arredond: Ok! I thought `version.py` was more standard looking at: https://stackoverflow.com/a/16084844/8159847

modules/s2/redshift/test/integration/test_VERSION.py
Jesus89: True that

modules/s2/redshift/lib/version.py
Jesus89: I have seen more `_version` in other projects, like numpy, pandas, geopandas

arredond:
Added missing functions, except for Hilbert ones

Jesus89:
OK, they can be added in the next PR

arredond:
I think that's all, if it looks ok we're good to merge

alex-graciano:
The overall PR LGTM but I added some minor fixes to be included. Nice job!

alex-graciano (changes requested):

arredond:
Nice catches @alex-graciano ! Forgot to update those. Should be resolved now

modules/s2/redshift/doc/ST_ASID_POLYFILL_BBOX.md
```diff
@@ -0,0 +1,42 @@
...
+**Description**
+
+Returns a SUPER containing an array of s2 cell IDs that cover a planar bounding box. Note that this is
+a compact coverage (polyfill), so the bounding box is covered with the least amount of cells 
+by using the largest cells possible.
+
+Two optional arguments can be passed with the minimum and maximum resolution for cogerage. If you desire a
+coverage at a single resolution level, simply set the maximum and minimum longitude to be of equal value.
+
+* `min_latitude`: `FLOAT` minimum latitude of the bounding box.
```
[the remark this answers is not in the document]
arredond: Its better to just use FLOAT. I started using NUMERIC because I though RS wouldn't automatically cast integers to float but it does so no need for them anymore. Updated in `sql/LONGLAT_ASID.sql`

modules/s2/redshift/test/unit/test_version.py
```diff
@@ -0,0 +1,13 @@
...
+
+def test_init():
+    assert s2Lib.__version__ == __version__
+    assert hasattr(s2Lib, 'get_cell_boundary')
+    assert hasattr(s2Lib, 'get_resolution')
+    assert hasattr(s2Lib, 'longlat_as_int64_id')
+    assert hasattr(s2Lib, 'int64_id_to_token')
+    assert hasattr(s2Lib, 'to_children')
+    assert hasattr(s2Lib, 'to_parent')
+    assert hasattr(s2Lib, 'uint64_repr_from_id')
```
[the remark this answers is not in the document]
arredond: Added, think I've got them all now

Jesus89 (approved):
Awesome job! LGTM

PR #167, opened 2021-09-14
Improve Python libraries deployment in Redshift

alex-graciano:
This will check if a library is already installed in Redshift so it won't be overwritten. For now just the name of the library is being checked since is the only information useful available in the Redshift `pg_library` table. Also, in this PR it's suggested to use two requirement files for module for the specific third-party libraries used in it and leave the common ones just for testing and system purposes.


Jesus89:
I found an issue with this approach: if the version number of the packages changes, since it only checks the name, it doesn't update the library. Making drop+create all the time would work but I'm going to try to detect both name+version.

Jesus89:
I've done several improvements:
- Install Python2 dependencies (instead of Python3)
- Check the library version (with an temporary UDF)
- Add a checksum system for the requirements.txt files

Jesus89 (approved):
LGTM. But we need to test it in both GNU/Linux and macOS

alex-graciano:
Testing on mac:
1. `make deploy` in line with the branch. Just the quadkey modules are installed. External Python modules are skipped.
2. Delete `requirements.txt.shasum` and `make deploy`. Try to install click and mercantile but both are already installed.
3. Change mercantile version. Error received `shasum: WARNING: 1 computed checksum did NOT match` and try `Installing mercantile==1.2.0 in redshift-cluster-1` which is overwritten in the cluster.

LGTM

PR #169, opened 2021-09-16
Feature/ch176393/add constructors module

vdelacruzb:
- Add constructors module.
- Make quadkey tilecover use python3 divisions. It's not affecting the polyfills in the tests but just in case.
- Update ST_TILEENVELOPE test in other warehouses. As we updated quadkey.ST_BOUNDARY it was affecting the ST_TILEENVELOPE test.The previous result didn't look good as it contained more than 4 points.

**Update** :heavy_check_mark: 
- Research why ST_MAKEELLIPSE result is slightly different from other warehouses. It might be because of any constant or decimals considerations. I attach a screenshot where this difference can be appreciated for a same input:
Captura de pantalla 2021-09-16 a las 12 12 04


alex-graciano:
LGTM. Regarding LICENSE and precision problems in `ST_MAKEELLIPSE`... Should they be solved before merging?

Jesus89:
We will review the LICENSEs in a future PR. Regarding `ST_MAKEELLIPSE` I'll take a look

Jesus89:
After increasing the precision, the result is better: blue - BigQuery, red - Redshift

PR #173, opened 2021-09-28
Redshift processing module

alex-graciano:
Implemented
- ST_POLYGONIZE (As alias of ST_MAKEPOLYGON)
- ST_DELAUNAYLINES (Using scipy.spatial)
- ST_DELAUNAYPOLYGONS (Using scipy.spatial)
- ST_VORONOILINES (Using scipy.spatial)
- ST_VORONOIPOLYGONS (Using scipy.spatial)

For Voronoi, Scipy does not return points of the diagram whose regions extents to the infinity so the computation of these lines should be done. In ST_VORONOILINES is not a big deal, but for polygons it becomes crucial since in Voronoi is important to know in which region a point falls. Now, the regions are extended by including the 4326 bbox into the diagram. Once Shapeli is installed a polygon clipping operation can be done.

In the figure can be seen that the polygons are enough large for including each point.

Captura de pantalla 2021-09-28 a las 16 26 21


alex-graciano:
I have realized a very simple way to compute the clipping with the BBOX in the `ST_VORONOIPOLYGONS` function. I'll ping you when developed @vdelacruzb @Jesus89.

modules/processing/redshift/doc/ST_DELAUNAYLINES.md
```diff
@@ -0,0 +1,22 @@
...
+
+**Description**
+
+Calculates the Delaunay triangulation of the points provided. A MultiLineString object is returned.
+
+* `points`: `GEOMETRY` MultiPoint input to the Delaunay triangulation.
+
+**Return type**
+
+`VARCHAR(MAX)`
```
vdelacruzb, line 15: We could return a `SUPER` here and also in voronoi functions by just doing a json_parse, so hopefully in the future (unless they implement new types) we don't have to readapt functions.

vdelacruzb (approved):
Approving for testing.

modules/processing/redshift/doc/ST_POLYGONIZE.md
```diff
@@ -0,0 +1,22 @@
+### ST_POLYGONIZE
```
vdelacruzb, line 1: BTW, this in BQ does the same that `ST_MAKEPOLYGON` right? Apart from accepting arrays do you know if there is any other reason to have it?

I'm also curious why the naming is not something more similar to the other naming, for instance: `ST_MAKEPOLYGON_ARR`.

vdelacruzb:
Nice job! On this I would mainly consider updating that return type.

vdelacruzb:
Also, we are not doing it in our own implementations (just in quadkey), but it's something created by CARTO here it would make sense to add unit tests (I know that it's a pain in the ass ^^').

modules/processing/redshift/doc/ST_DELAUNAYLINES.md
vdelacruzb: They sent us a solution on how to use `SUPER` types so it's the way to go. I'm still testing as they say that we have to add extra paths if we want them to contain schemas.

alex-graciano:
Version with bounding box clipping:
Captura de pantalla 2021-10-04 a las 8 51 53

modules/processing/redshift/doc/ST_POLYGONIZE.md
alex-graciano: I think it was added because it appears in PostGIS with that name, however we can delete this if @Jesus89 agrees. For now, it's a simple alias. Let's see in QA

alex-graciano:
The precision is set to 10 in order to produce at least medium diagrams. If precision is sets to 15 some tests that used to work nice now raise an exception `Value too long for character type`

PR #174, opened 2021-09-29
fix setool parameters

vdelacruzb:
The parameters where being concatenated wrongly for all the warehouses example:
INPUT:
```
param1,param2
INT,FLOAT
```

Was producing something like:
`param1 param2, INT FLOAT`

Instead of:
`param1 INT, param2 FLOAT`

### DIFF
--- a/common/redshift/Makefile
+++ b/common/redshift/Makefile
@@ -88,7 +88,7 @@ clean-deploy: check
 	$(MAKE) share-remove
 	$(MAKE) schema-remove || ((sleep 5 && $(MAKE) schema-remove) || exit 1)
 
-storage-upload: build venv3
+storage-upload: build venv2 venv3
 	for f in $(wildcard dist/*.zip); do \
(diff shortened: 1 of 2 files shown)


common/redshift/Makefile
```diff
@@ -88,7 +88,7 @@ clean-deploy: check
 	$(MAKE) share-remove
 	$(MAKE) schema-remove || ((sleep 5 && $(MAKE) schema-remove) || exit 1)
 
-storage-upload: build venv3
+storage-upload: build venv2 venv3
```
Jesus89, line 91: I think venv3 is not needed here.

vdelacruzb: You are right, I just removed it and it works locally. It's weird though since previously it was mostly required for having wheel installed in the environment. Let's see if it works in the PR.

Jesus89:
Hey, I think this is because now the tool uses `;` as separator instead of `,` (because of complex types). I would review the form for the input params.

PR #177, opened 2021-10-06
fix constructors precision

vdelacruzb:
I've moved the loads and dumps to the lib to make the sql cleaner.

The integration tests results are changing because of the latests decimals and also because, we were using str() before and geojson.dumps is bringing the geom type to the front.


modules/constructors/redshift/lib/__init__.py
```diff
@@ -1,15 +1,8 @@
 # Copyright (c) 2021, CARTO
 
 from ._version import __version__  # noqa
+from ellipse import ellipse
```
alex-graciano, line 4: Use better relative routes, i.e. .ellipse,...

alex-graciano (changes requested):
Apart from the paths it LGTM

alex-graciano (approved):
Nice job!

PR #182, opened 2021-10-29
Define snowflake UDFs as immutable

jgoizueta:
If a function value depends only on it's argument values (so most functions except things link random stuff, user subscriptions) then it should be declared IMMUTABLE on Snowflake, since the default is VOLATILE and can get evaluated unnecessarily.

The Delaunay function are failing the test, both locally and CI for some reason.


alex-graciano:
It seems that some rows, although valid, are swapped. I think Snowflake should have change something internal. Fixing it.

Jesus89:
Could you run the tests locally several times? Maybe the function is not deterministic. Or maybe it's just a difference when using IMMUTABLE vs VOLATILE

alex-graciano:
> Could you run the tests locally several times? Maybe the function is not deterministic. Or maybe it's just a difference when using IMMUTABLE vs VOLATILE

I've executed the test with the old development version that I had in my account and with the `INMUTABLE` tag and both returned the same, it seems that is not related with the new changes.

alex-graciano (approved):
Test passing now

PR #185, opened 2021-11-09
Cache libraries in all snowflake udfs

### LINKED ISSUE #184: Snowflake Performance
On Snowflake, the performance of certain functions from here are very poor.

I believe the fix is simple, and it uses caching of  libraries across function calls:

https://docs.snowflake.com/en/developer-guide/udf/javascript/udf-javascript-introduction.html#global-state

For instance, for `LONGLAT_AS_H3`, you can modify the existing code with some `setup` function like so:

```SQL
CREATE OR REPLACE FUNCTION @@SF_PREFIX@@h3._LONGLAT_ASH3
(longitude DOUBLE, latitude DOUBLE, resolution DOUBLE)
RETURNS STRING
LANGUAGE JAVASCRIPT
IMMUTABLE
AS $$
    function setup() {
        @@SF_LIBRARY_LONGLAT_ASH3@@
        h3LibGlobal = h3Lib;
    }
    if (typeof(setup_done) === "undefined") {
        setup();
        setup_done = true;
    }
    if (LONGITUDE == null || LATITUDE == null || RESOLUTION == null) {
        return null;
    }
    const index = h3LibGlobal.geoToH3(Number(LATITUDE), Number(LONGITUDE), Number(RESOLUTION));
    return index;
$$;
```

On my informal benchmarking, adding this caching allows this function to be called 1 Million times in 3-4 seconds, while without caching, it never completes (> 5 minutes and counting).

Have you come across similar performance problems? Do you see any reason **not** to adopt this pattern across all Snowflake calls?

vdelacruzb:
fixes #184


modules/constructors/snowflake/test/integration/ST_BEZIERSPLINE.test.js
```diff
@@ -8,7 +8,7 @@ test('ST_BEZIERSPLINE should work', async () => {
     const rows = await runQuery(query);
     expect(rows.length).toEqual(1);
     expect(JSON.stringify(rows[0].BEZIERSPLINE1)).toEqual('{"coordinates":[[121.025390625,-22.91792293614603],[122.40878796386718,-22.04579264902266],[125.672291015625,-20.380801787862705],[129.4858663330078,-19.350050352981388],[132.704296875,-20.01569558887108],[135.25804138183594,-21.75113289066844],[137.26212890625,-24.05282839437402],[138.7404580078125,-26.51113113177313],[139.065328125,-29.074453813709248],[138.68492871093753,-31.163421738140798]],"type":"LineString"}');
-    expect(JSON.stringify(rows[0].BEZIERSPLINE2)).toEqual('{"coordinates":[[-6,-0.5],[-3.6649305555555554,0.4259259259259258],[-1.0611111111111116,-0.24074074074074053],[1.5,0],[4.06111111111111,0.24074074074074117],[6.664930555555557,-0.42592592592592565]],"type":"LineString"}');
+    expect(JSON.stringify(rows[0].BEZIERSPLINE2)).toEqual('{"coordinates":[[-6,-0.5],[-3.6649305555555554,0.4259259259259258],[-1.0611111111111116,-0.24074074074074053],[1.5,0],[4.06111111111111,0.24074074074074117],[6.664930555555557,-0.42592592592592565],[9,0.5]],"type":"LineString"}');
```
jgoizueta, line 11: I guess this is unrelated, but do you know what's going on there?

jgoizueta (approved):
LGTM 
🏎️💨

Jesus89 (changes requested):
Nice trick!

However, I would revisit the implementation to use only `@@SF_LIBRARY_CONTENT@@` and keep using `{MODULE}Lib`.

This is a more compact solution:

```js
if (typeof(h3Lib) === "undefined") {
    var _h3Lib=function(){...  // lib code
    h3Lib = _h3Lib
}
```

We rename `h3Lib` to `_h3Lib`, so we can use `h3Lib` as global (by hoisting).

My proposal is to replace `@@SF_LIBRARY_CONTENT@@` with this logic so it's automatically included in all the functions and use `{MODULE}Lib` for the calls.

modules/constructors/snowflake/test/integration/ST_BEZIERSPLINE.test.js
vdelacruzb: Yes this is a totally unrelated topic. This test was crashing before the changes and we didn't notice. ST_BEZIERSPLINE was returning a line that wasn't reaching the end of the input line.
Captura de pantalla 2021-11-10 a las 9 36 10
But now it does.
Captura de pantalla 2021-11-10 a las 9 36 57

vdelacruzb:
I was verifying that states are kept per function and not between functions since I was afraid using the same variable `h3Lib` for different UDFs could cause interferences as each one contains its own implementation. It looks ok, states should be per function as the docs say.

Jesus89 (approved):
LGTM!

Clean implementation using the config.

PR #192, opened 2021-11-24
minor changes to sql reference

martejpad:


vdelacruzb (approved):
Nice, LGTM. I see that most of this doc was equal to BQ and SF. We might want to update the other warehouses as well.

PR #193, opened 2021-11-30
Bigquery naming refactor

vdelacruzb:


common/bigquery/Makefile
```diff
@@ -1,29 +1,40 @@
...
 
+CLOUD = bigquery
+
 SED = sed
 GSUTIL = gsutil
 BQ = bq --location=$(BQ_REGION) --project_id=$(BQ_PROJECT)
 
+BQ_MODULE_LABEL ?= spatial_extension_module:core
+BQ_DEPLOY_DATASET = $(BQ_DATASET_PREFIX)carto
 BQ_DATASET = $(BQ_DATASET_PREFIX)$(MODULE)
```
jgoizueta, line 11: Could we rename this? (I assume it won't be used as a dataset name, right?) perhaps BQ_DEPLOY_MODULE ?

common/bigquery/Makefile
```diff
@@ -102,30 +148,47 @@ storage-remove:
 	fi
 
 dataset-create:
-	$(BQ) show $(BQ_DATASET) 2>/dev/null 1>/dev/null || \
-		$(BQ) mk -d --description "$(BQ_DATASET) module" -label $(BQ_MODULE_CORE_LABEL) $(BQ_DATASET)
+	$(BQ) show $(BQ_DEPLOY_DATASET) 2>/dev/null 1>/dev/null || \
+		$(BQ) mk -d --description "$(BQ_DEPLOY_DATASET) module" -label $(BQ_MODULE_LABEL) $(BQ_DEPLOY_DATASET)
+ifdef BQ_CREATE_SPATIAL_CATALOG
```
jgoizueta, line 153: We could we replace this by a more generic mechanism, like a `BQ_POST_DATASET_CREATE` pointing to a shell script that we define in the data module, then we do something like this here:
```
ifdef BQ_POST_DATASET_CREATE
  BQ_POST_DATASET_CREATE
endif
```

common/bigquery/geonames/01_create_names.sql
```diff
@@ -0,0 +1,118 @@
+CREATE OR REPLACE TABLE `@@BQ_PREFIX@@carto.names` (
```
jgoizueta, line 1: We'll remove this (it's handled by DO now)

common/bigquery/Makefile
Jesus89: BQ_DEPLOY_DATASET is the dataset where the toolbox is installed.

Jesus89: I have changed `$(BQ_DATASET_PREFIX)$(MODULE)` to be used directly in the bucket variable

common/bigquery/Makefile
Jesus89: Good point. I have checked that a strategy using a Makefile rule `post-dataset-create:` works

common/bigquery/geonames/01_create_names.sql
Jesus89: :fire:

Jesus89 (approved):

PR #194, opened 2021-12-02
Adding carto-un-na-ne1 project

jvillarf:
https://app.shortcut.com/cartoteam/story/193828/telus-create-analytics-toolbox-project-in-northamerica-northeast1

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -69,7 +69,7 @@ jobs:
             project: carto-un-eu
             bucket: gs://spatialextension_un/
             group: spatialextension_users_un@cartodb.com
-          - region: na-ne1
+          - region: northamerica-northeast1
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-production.yml
```diff
@@ -69,6 +69,10 @@ jobs:
             project: carto-un-eu
             bucket: gs://spatialextension_un/
             group: spatialextension_users_un@cartodb.com
+          - region: na-ne1
```
Jesus89, line 72: I think you need to put here the full name for the region: `northamerica-northeast1`

jvillarf: You are right: https://github.com/CartoDB/carto-spatial-extension/pull/194/commits/782dca4fca1f0aa35a68e4526075983bd672d041

Jesus89 (approved):

PR #196, opened 2021-12-14
Split Redshift release into modules.sql and libraries.sql

vdelacruzb:
This PR stores library creation queries and UDFs on different sql as well as removing library duplicates.


Jesus89:
This was already done and merged in master. Taking the tip of "sort -u"

PR #198, opened 2021-12-16
Nested udfs error in redshift

alex-graciano:
This refactor saves us from the extra nesting that produces the Redshift error in the `transformations` module:

```
SQL Error [500310] [XX000]: [Amazon](500310) Invalid operation: Nested UDFs are not supported.;
```

https://app.shortcut.com/cartoteam/story/198004/nested-udfs-error-in-redshift


modules/transformations/redshift/lib/helper.py
```diff
@@ -103,3 +105,45 @@ def calculate_final_bearing(start, end):
     bear = bearing(end, start)
     bear = (bear + 180) % 360
     return bear
+
+
+def geom_from_geojson(geom):
```
vdelacruzb, line 110: I would rename this functions to avoid confusions `geom_from_geojson`-> `wkt_from_geojson`.

vdelacruzb (commented):
I would do the rename that I mentioned and then should be good to go.
Makes me worried that from other modules, the same nesting bug could still be happening as we are not doing stuffs internally but calling `__ST_GEOMFROMGEOJSON`, hope the oficial function arrives soon.

alex-graciano:
> __ST_GEOMFROMGEOJSON

vdelacruzb (approved):
LGTM

Jesus89 (commented):
Could you add a test to verify this is fixed? Thanks

alex-graciano:
I was working on one but I cannot imagine why the Python connector was not let me `CREATE TABLE... AS` from a query. I'll take a look

Jesus89 (approved):

PR #202, opened 2021-12-22
Review installation packages

Jesus89:
- Generate zip files
- Release on merge
- Load global version from file


.github/workflows/release-production.yml
```diff
@@ -1,12 +1,10 @@
 name: Release Production
 
 on:
+  push:
+    branches:
+      - master
   workflow_dispatch:
-    inputs:
```
vdelacruzb, line 5: I guess we are removing this because it will be triggered automatically when pushing. I think it was pretty useful to select warehouses and would maybe add it to `deploy-production` in the future as optional. If `github.event.inputs.cloud` not defined deploy all, and if not we deploy specific warehouses. Some days ago I was triggering few times the deployment of everything and just needed to do it for Snowflake.

vdelacruzb (approved):
Nice refactor. LGTM!

.github/workflows/release-production.yml
Jesus89: OK, we can include it for both: deploy-production, release-production

PR #205, opened 2022-01-12
Add carto-un-au-se1 (Australia) and carto-un-as-ne1 (Asia) projects

miguelangelmorenochacon:


Jesus89 (changes requested):
The region must be the full name in Google Cloud

Jesus89 (approved):

PR #208, opened 2022-01-20
add H3_RESOLUTION function

arredond:
Adds a new `H3_RESOLUTION` to get the resolution of an H3 Cell ID as an INTEGER.
Adds a new `H3_CENTER` to get the center of an H3 Cell.

PR #204, opened 2022-01-05
fix wrap error at the pole for siblings and kring quadkey, convert in pure sql, add QUADINT_CENTER

francois-baptiste:
* fix wrap error at the pole for siblings
  * QUADINT_KRING
  * QUADINT_KRING_DISTANCES
  * QUADINT_SIBLING
* convert in pure sql
  * QUADINT_KRING
  * QUADINT_KRING_DISTANCES
  * QUADINT_SIBLING
* add QUADINT_CENTER
* add QUADINT_RESOLUTION

PR #209, opened 2022-01-24
add QUADINT_RESOLUTION function

arredond:
Similar to #208 , add a function to get the resolution of a quadint.

Under the hood this simply converts the quadint to a quadkey and gets its length. Pretty trivial but the average user probably wouldn't know how to do it on their own.


francois-baptiste:
@arredond Can you do this in pure SQL in my PR #204 ?

francois-baptiste:
BTW  the resolution is the z in the QUADINT_TOZXY function:
```sql
SELECT QUADINT_TOZXY(4834).z
-- returns 2
```
I don't think there is a need to expose a new JS function in the lib.

arredond:
Already merged #211 into #204 , closing this PR

PR #213, opened 2022-02-02
Use dotenv file

Jesus89:
Use `.env` file for the environment variables. These are loaded in each Makefile script.


.github/workflows/ci.yml
```diff
@@ -127,11 +127,11 @@ jobs:
...
     env:
       SNOWSQL_VERSION: 1.2.13
-      SNOWSQL: ~/snowflake/snowsql
+      SNOWSQL_BIN: ~/snowflake/snowsql
       SNOWSQL_DEST: ~/snowflake
-      SNOWSQL_ACCOUNT: ${{ secrets.SF_ACCOUNT }}
-      SNOWSQL_USER: ${{ secrets.SF_USERNAME }}
-      SNOWSQL_PWD: ${{ secrets.SF_PASSWORD }}
+      SF_ACCOUNT: ${{ secrets.SF_ACCOUNT }}
+      SF_USER: ${{ secrets.SF_USERNAME }}
```
vdelacruzb, line 133: Now, wish that we would have called the secret SF_USER but it's a minor thing.

.env.template
```diff
@@ -0,0 +1,41 @@
+# This is a sample .env file for use in local development.
+# Duplicate this file as .env in the root of the project
+# and update the environment variables to match your
+# desired config
```
vdelacruzb, line 4: I guess we will assume that most of people working on this will know that $ has to be scaped in .env files. If not we could had a comment

vdelacruzb (approved):
Nice refactor. Go for it!

PR #221, opened 2022-02-24
Deploy to CI Datawarehouses

miguelangelmorenochacon:
Related to
- https://github.com/CartoDB/cloud-native/pull/4878


Jesus89:
Hi, sorry for the delay.

In order to pass the role and warehouse, the file `common/snowflake/Makefile` must be modified to pass the params with -r and -w respectively to `snowsql`. I'll update this today so we can merge.

Jesus89:
Hi, we have done a change in the GH Action scripts, which is not compatible with this approach. Moreover, I would apply only this change in the AT repo, and with a job `deploy-ci` inside bigquery, snowflake, redshift, and postgres scripts. I'll ping you for more info.

PR #222, opened 2022-03-03
Add utils for unit tests (which use python2)

jgoizueta:
Add norm_sql function to aid in unit tests.
Note that unit tests uses venv2, so we can't use `test_utils` in those because it requires libraries only present in venv3.


alex-graciano (approved):
LGTM

PR #223, opened 2022-03-08
St tileenvelope returning json varchar in redshift

vdelacruzb:
We are still doing str() castings in more of our functions like polyfills. At some point we will have to replace all of them by json.dumps.

Didn't make the change still to focus on the tiler.


alex-graciano (approved):
LGTM

PR #230, opened 2022-03-23
Add ST_BUFFER in Snowflake

Jesus89:
This PR adds the `ST_BUFFER` function in the `transformations` module for Snowflake.

The implementation reduces the required JS code to 230 kB so it fits in a JS UDF.


modules/transformations/snowflake/test/integration/ST_BUFFER.test.js
```diff
@@ -0,0 +1,20 @@
...
+    expect(rows[0].BUFFER).toEqual(pointFixturesOut.value);
+});
+
+test('ST_BUFFER should work with segments', async () => {
+    const featureJSON = JSON.stringify(pointFixturesIn.geom.geometry);
+    const query = `SELECT ST_ASTEXT(ST_BUFFER(TO_GEOGRAPHY('${featureJSON}'), 1000, 8)) as buffer;`;
+    const rows = await runQuery(query);
+    expect(rows.length).toEqual(1);
+    expect(rows[0].BUFFER).toEqual(pointFixturesOut.value);
+});
```
vdelacruzb, line 20: Maybe we can add a test for default values as we do in other turf UDFs.

modules/transformations/snowflake/sql/ST_BUFFER.sql
```diff
@@ -0,0 +1,42 @@
...
+IMMUTABLE
+AS $$
+    if (!GEOJSON || DISTANCE == null || SEGMENTS == null) {
+        return null;
+    }
+
+    const options = {
+        units: 'meters',
+        steps: Number(SEGMENTS)
+    };
```
vdelacruzb, line 18: Is not something meaningful at all. But in the rest of UDFs we are declaring options after @@SF_LIBRARY@@

vdelacruzb (approved):
LGTM. Please remember to update the common version before submitting

modules/transformations/snowflake/test/integration/ST_BUFFER.test.js
Jesus89: The default is covered. I'll change the test passing the segment to a value different than the default one

martejpad (approved):

PR #232, opened 2022-03-28
Alter binary representation of quadints

mhconradt:
This interleaves the binary representation of quadints, inspired by: https://github.com/CartoDB/python-quadkey. This representation uses the least-significant bits of the quadint to store the level index. 
This is backwards incompatible with the existing quadint representation and thus breaks some of the tests, but I wanted to share with the community.


Jesus89:
Hi, thanks for the contribution.

This is a big (breaking?) change we need to discuss with the product managers (speed vs clustering).

Jesus89:
Hi, we have introduced a new encoding called "quadbin" that will replace current quadint functions. It included the interleaved xy and the z.

It's implemented in pure SQL. Feel free to take a look and test it: https://github.com/CartoDB/analytics-toolbox-core/pull/255.

PR #241, opened 2022-04-21
Add util to sort expected rows and round numeric values

jgoizueta:
This is handy for comparing result rows in integration tests, given that order may not be specified and float number accuracy may vary.


vdelacruzb (approved):
LGTM

PR #243, opened 2022-04-26
Fix Quadkey functions not consistent with other warehouses

vdelacruzb:
I notice that apart from `QUADINT_ASQUADKEY`, `QUADINT_ASZXY` was not consistent in the naming.


Jesus89 (approved):
LGTM

PR #246, opened 2022-05-11
Reimplement BQ quadkey module in SQL

jgoizueta:


Jesus89:
This is gold :1st_place_medal: Thanks for that leapfrog!

It has been the seed for the new quadbin index: https://github.com/CartoDB/analytics-toolbox-core/pull/255, which will replace quadkey/quadint.

PR #247, opened 2022-05-16
Use `?=` to set the dataset name in BigQuery Makefile

osuketh:
It allows users to set the name of the BigQuery Dataset.


osuketh:
I just noticed the names of datasets are also used in the SQL files which create functions, so I closed this PR.

PR #253, opened 2022-06-21
Add quadbin module for Redshift

vdelacruzb:


vdelacruzb:
WIP:

- Fix unit tests.
- Fix integration tests.
- Add QUADBIN_ISVALID, QUADBIN_CENTER and QUADBIN_RESOLUTION.
- Try to use more SQL implementations.
- Replace quadbin codification for proposal 3?

vdelacruzb:
I have used the same implementation for toparent that we have in BQ. I could make it fully SQL what should be faster but I will have to remove the exceptions regarding wrong resolution.

vdelacruzb:
Also `QUADBIN_BOUNDARY` should be updated once Redshift support `ST_GEOMFROMGEOJSON`

modules/quadbin/redshift/lib/tilecover/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+            t = tiles[i]
+            tile_hash[to_id(t[0], t[1], t[2])] = True
+
+        return merge_tiles(tile_hash, tiles, limits)
+
+    append_hash_tiles(tile_hash, tiles)
+    return tiles
+
+
+def merge_tiles(tile_hash, tiles, limits):
```
jgoizueta, line 54: So the RS polyfills compacts the tiles?
If so shouldn't that be optional? (and have by default the same behaviour we have in the other providers)

modules/quadbin/redshift/sql/QUADBIN_BOUNDARY.sql
```diff
@@ -0,0 +1,18 @@
+----------------------------
+-- Copyright (C) 2021 CARTO
+----------------------------
+
+CREATE OR REPLACE FUNCTION @@RS_PREFIX@@carto.QUADBIN_BOUNDARY
+(quadbin BIGINT)
+RETURNS VARCHAR(MAX)
```
jgoizueta, line 7: Refresh my mind: we don't return a GEOMETRY because Redshift lacks GeoJSON to Geometry conversion?

jgoizueta: OK, I just saw you comment about it

vdelacruzb: It's because RS doesn't have still `ST_GEOMFROMGEOJSON`. Actually as it's going to be a square we could change the code to return a WKT, until the function arrives in July.

vdelacruzb:
BTW I have removed the error handling from `QUADBIN_TOPARENT` because it was behaving bad when using the function with tables. I have tried the previous python solution but for the table `derived_spatialfeatures_ukr_quadgrid15_v1_yearly_v2` it took around 17 secs while it took 3.1 secs with the SQL version.

modules/quadbin/redshift/sql/QUADBIN_BOUNDARY.sql
jgoizueta: It would be nice to have the GEOMETRY already using WKT

modules/quadbin/redshift/sql/QUADBIN_POLYFILL.sql
```diff
@@ -0,0 +1,42 @@
...
+    return json.dumps(quadbins)
+$$ LANGUAGE plpythonu;
+
+CREATE OR REPLACE FUNCTION @@RS_PREFIX@@carto.QUADBIN_POLYFILL
+(GEOMETRY, INT)
+-- (geo, resolution)
+RETURNS SUPER
+STABLE
+AS $$
+    SELECT json_parse(@@RS_PREFIX@@carto.__QUADBIN_POLYFILL(ST_ASGEOJSON($1)::VARCHAR(MAX), $2))
```
jgoizueta, line 41: Does RS supports SRIDs in the GEOMETRY? For PG I took the trouble to support any SRID using ST_TRANSFORM.... but we could also specify that input geometries must be 4326

vdelacruzb: Yes, RS supports different SRIDs I'll be adding `ST_SetSRID` to these kind of cases.

modules/quadbin/redshift/lib/tilecover/__init__.py
vdelacruzb: Has we have discussed. It looks like we are not entering that merge_tiles function because of setting both min and max zoom as `resolution`.

modules/quadbin/redshift/sql/QUADBIN_POLYFILL.sql
vdelacruzb: Updated in latest commit

modules/quadkey/redshift/sql/QUADINT_FROMGEOGPOINT.sql
```diff
@@ -8,5 +8,5 @@ CREATE OR REPLACE FUNCTION @@RS_PREFIX@@carto.QUADINT_FROMGEOGPOINT
 RETURNS BIGINT
 STABLE
 AS $$
-    SELECT @@RS_PREFIX@@carto.QUADINT_FROMLONGLAT(ST_X($1), ST_Y($1), $2)
+    SELECT @@RS_PREFIX@@carto.QUADINT_FROMLONGLAT(ST_X(ST_SetSRID($1, 4326)), ST_Y(ST_SetSRID($1, 4326)), $2)
```
jgoizueta, line 11: But ST_SetSRID blindly sets the SRID. If the geometry is originally in other reference we're getting the wrong coordinates. We should use ST_TRANSFORM to compute the correct 4326 coordinates. But it fails if the geometry doesn't have a SRID (0). What I ended up doing in PG is:
`CASE WHEN ST_SRID($1) = 0 THEN ST_SetSRID($1, 4326) ELSE ST_Transform($1, 4326) END`

vdelacruzb: Updated

modules/quadbin/redshift/sql/QUADBIN_BOUNDARY.sql
vdelacruzb: Updated

vdelacruzb:
BTW in all the clouds we are doing stuffs like:
`SELECT carto.QUADBIN_FROMLONGLAT(40.4168, -3.7038, 4);`

Madrid coordinates would be:
`SELECT carto.QUADBIN_FROMLONGLAT(-3.7038, 40.4168, 4);`
Anyways... Mombasa seems also a nice place to be.

jgoizueta (approved):

PR #256, opened 2022-06-22
Add quadbin module for Snowflake (don't merge)

volaya:
This PR adds quadbin support to SF, but has some issues:

- The QUADBIN_TOCHILDREN function is not functional. I tried to put the code in a single function, but it was throwing `Unsupported subquery type cannot be evaluated`. I split the code in several functions, which are functional when used manually (that is, running one and then manually taking its result and calling the next one). That works fine, but when put together, it systematically throws `SQL execution internal error: Processing aborted due to error 300002:834137686`. I couldn't find much about what that means.

I think I have seen this type of problem before, specially when calling a UDF from another one involving bitshift operations, but I cannot figure out a way of solving it.

I suspect the issue might be related to the _GENERATE_RANGE function that I had to add to replace the BQ GENERATE_ARRAY function, since I can reproduce the error using that function in another context. However it is working fine in other functions in this PR that use it.

I have left the function in the PR, so it can be reviewed, but maybe it can be removed and the PR merged without it. We can work on that later

- The QUADINT_BOUNDARY function contains some nasty ugly code. I was having issue with a nicer and more standard code (Snowflake died systematically when calling the corresponding CREATE FUNCTION. It kept computing for a long time and I was forced to stop it), and adding a chain of CTEs was the only workaround I could find for creating the LineString required in that function. No idea why that happens. If anyone has a nicer way of doing that, it would be good to change it, for the sake of readability.


volaya:
QUADBIN_TOCHILDREN should be fixed now (Thanks @jgoizueta!!). Tests are failing because all queries involving the new quadbin functions are returning undefined. I will investigate to see what's happening.

volaya:
Test are fixed and they are all passing now.

modules/quadbin/snowflake/sql/QUADBIN_TOZXY.sql
```diff
@@ -0,0 +1,63 @@
...
+(quadbin INT)
+RETURNS OBJECT
+AS $$
+WITH
+__interleaved AS (
+  SELECT
+    BITAND(BITSHIFTRIGHT(quadbin, 59), 7) AS mode,
+    BITAND(BITSHIFTRIGHT(quadbin, 57), 3) AS extra,
+    BITAND(BITSHIFTRIGHT(quadbin, 52), 31) AS z,
+    BITSHIFTLEFT(BITAND(quadbin, TO_NUMBER('fffffffffffffff', 'XXXXXXXXXXXXXXXX')), 12) AS q
```
jgoizueta, line 11: I would avoid the TO_NUMBER calls and use contants; you can get the decimal values from the Postgres PR

modules/quadbin/bigquery/README.md
```diff
@@ -1,3 +1,3 @@
 # Quadbin module for BigQuery
 
-You can learn more about quadbins in the [Overview section](/analytics-toolbox-bigquery/overview/spatial-indexes/#quadbin) of the documentation.
\ No newline at end of file
+You can learn more about quadbins in the [Overview section](/analytics-toolbox-bigquery/overview/spatial-indexes/#quadbin) of the documentation.
```
vdelacruzb, line 3: I guess this is a branch of `feature/sc-238749/add-quadbin-module-all-clouds-bigquery` and latest changes were not taken so it consider the files in bigquery folder as added code.

Jesus89:
Superseded by https://github.com/CartoDB/analytics-toolbox-core/pull/258

PR #258, opened 2022-06-23
Add quadbin module for Snowflake

volaya:


modules/quadbin/snowflake/README.md
```diff
@@ -0,0 +1,2 @@
+# Quadbin module for Snowflake
+TODO: add module description.
```
vdelacruzb, line 2:
```suggestion
# Quadbin module for Snowflake

You can learn more about quadbins in the [Overview section](/analytics-toolbox-snowflake/overview/spatial-indexes/#quadbin) of the documentation.
```

modules/quadbin/snowflake/doc/QUADBIN_STRING_TOINT.md
```diff
@@ -0,0 +1,24 @@
+### QUADBIN_STRING_TOINT
```
vdelacruzb, line 1: This can be removed as now is a private UDF

modules/quadbin/snowflake/doc/QUADBIN_INT_TOSTRING.md
```diff
@@ -0,0 +1,24 @@
+### QUADBIN_INT_TOSTRING
```
vdelacruzb, line 1: This can be removed as now is a private UDF

modules/quadbin/snowflake/sql/QUADBIN_INT_TOSTRING.sql
```diff
@@ -0,0 +1,10 @@
+----------------------------
+-- Copyright (C) 2022 CARTO
+----------------------------
+
+CREATE OR REPLACE FUNCTION QUADBIN_INT_TOSTRING
```
vdelacruzb, line 5:
```suggestion
CREATE OR REPLACE FUNCTION _QUADBIN_INT_TOSTRING
```

vdelacruzb: We should rename the file name as well

modules/quadbin/snowflake/sql/QUADBIN_STRING_TOINT.sql
```diff
@@ -0,0 +1,10 @@
+----------------------------
+-- Copyright (C) 2022 CARTO
+----------------------------
+
+CREATE OR REPLACE FUNCTION QUADBIN_STRING_TOINT
```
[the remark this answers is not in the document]
vdelacruzb: We should rename the file name as well

modules/quadbin/snowflake/sql/_QUADBIN_SIBLING.sql
```diff
@@ -0,0 +1,12 @@
+----------------------------
```
vdelacruzb, line 1: Maybe we can just group this UDF inside QUADBINT_SIBLING.sql

modules/quadbin/snowflake/sql/_ZXY_QUADBIN_SIBLING.sql
```diff
@@ -0,0 +1,18 @@
+----------------------------
```
vdelacruzb, line 1: This could go to QUADBIN_SIBLING.sql as well

modules/quadbin/snowflake/test/integration/QUADBIN_STRING_TOINT.test.js
```diff
@@ -0,0 +1,8 @@
+const { runQuery } = require('../../../../../common/snowflake/test-utils');
```
vdelacruzb, line 1: This test can now be removed as the UDF is now private.

modules/quadbin/snowflake/quadbin.js
```diff
@@ -0,0 +1,68 @@
+function tileToQuadbin (tile) {
```
vdelacruzb, line 1: Not sure if you want to keep this file. It seems like using the old codification. Also there are few console.logs at the end

modules/quadbin/snowflake/doc/_INTRO.md
```diff
@@ -0,0 +1,5 @@
+## quadbin
+
+<div class="badges"><div class="core"></div></div>
+
+TODO.
```
vdelacruzb, line 5:
```suggestion
## quadbin

<div class="badges"><div class="core"></div></div>

You can learn more about quadbins in the [Overview section](/analytics-toolbox-snowflake/overview/spatial-indexes/#quadbin) of the documentation.
```

modules/quadbin/snowflake/doc/QUADBIN_TOZXY.md
```diff
@@ -0,0 +1,25 @@
...
+
+{{% bannerNote type="code" %}}
+carto.QUADBIN_TOZXY(quadbin)
+{{%/ bannerNote %}}
+
+**Description**
+
+Returns the zoom level `z` and coordinates `x`, `y` for a given quadbin.
+
+* `quadbin`: `INT` quadbin we want to extract tile information from.
```
vdelacruzb, line 11: For `quadint` we were using the type BIGINT for everything quadint related. This is the same case. But it seems that they are synonyms so we can keep it if you we are not losing precision

modules/quadbin/snowflake/doc/QUADBIN_TOCHILDREN.md
```diff
@@ -0,0 +1,28 @@
...
+**Description**
+
+Returns an array with the children quadbins of a given quadbin for a specific resolution. A children quadbin is a quadbin of higher level of detail that is contained by the current quadbin. Each quadbin has four children by definition.
+
+* `quadbin`: `INT` quadbin to get the children from.
+* `resolution`: `INT` resolution of the desired children.
+
+**Return type**
+
+`ARRAY<INT>`
```
vdelacruzb, line 16: In Snowflake unfortunately we only have the type Array not ARRAY<INT>. I think it's a good idea to specify the type inside the array but in the other docs we are just writing ARRAY. It's a pain in the as also trying to convert elements from the array to other types inside a UDF as we were getting correlated subqueries issues all the time.

modules/quadbin/snowflake/doc/QUADBIN_POLYFILL.md
```diff
@@ -0,0 +1,32 @@
...
+**Return type**
+
+`ARRAY<INT>`
+
+{{% customSelector %}}
+**Example**
+{{%/ customSelector %}}
+
+```sql
+SELECT carto.QUADBIN_POLYFILL(
```
vdelacruzb, line 23: For this example and the test we are doing the next in other warehouses:
```
carto.QUADBIN_POLYFILL(
    ST_GEOGFROMTEXT('POLYGON ((-3.71219873428345 40.413365349070865, -3.7144088745117 40.40965661286395, -3.70659828186035 40.409525904775634, -3.71219873428345 40.413365349070865))'),
    17);
```

So we avoid all the points, makeline and makepolygon

modules/quadbin/snowflake/test/integration/QUADBIN_INT_TOSTRING.test.js
```diff
@@ -0,0 +1,8 @@
+const { runQuery } = require('../../../../../common/snowflake/test-utils');
```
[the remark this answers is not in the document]
jgoizueta: I would keep the test, since it's not like other internal functions that are used from public functions and tested that way.

modules/quadbin/snowflake/sql/QUADBIN_BBOX.sql
```diff
@@ -0,0 +1,148 @@
+-- This method should call QUADBIN_TOZXY, but that causes SnowFlake to crash,
+-- so we are computing the ZXY here, in-place
+CREATE OR REPLACE FUNCTION QUADBIN_BBOX
+(quadbin INT)
+RETURNS ARRAY
+AS $$
+WITH __interleaved AS (
```
jgoizueta, line 7: Can you add a special case to return NULL when quadbin is NULL? We're doing that in the other providers. (I think I was the one that dropped the check when I looked for a workaround for the SF crash)

modules/quadbin/snowflake/sql/QUADBIN_BBOX.sql
```diff
@@ -0,0 +1,148 @@
...
+    SELECT
+        BITAND(BITSHIFTRIGHT(quadbin, 59), 7) AS mode,
+        BITAND(BITSHIFTRIGHT(quadbin, 57), 3) AS extra,
+        BITAND(BITSHIFTRIGHT(quadbin, 52), 31) AS z,
+        BITSHIFTLEFT(
+            BITAND(
+                quadbin,
+                TO_NUMBER('fffffffffffffff', 'XXXXXXXXXXXXXXXX')
```
jgoizueta, line 15: I guess this is very minor, but I would use the corresponding decimal constant (like 1152921504606846975 in this case) to remove a few TO_NUMBER calls (maybe SF is smart enough to optimize that away, but just in case...)

modules/quadbin/snowflake/sql/QUADBIN_BOUNDARY.sql
```diff
@@ -0,0 +1,70 @@
+----------------------------
+-- Copyright (C) 2022 CARTO
+----------------------------
+CREATE
+OR REPLACE FUNCTION QUADBIN_BOUNDARY(quadbin INT) 
+RETURNS GEOGRAPHY 
+AS $$ 
+CASE
+    WHEN quadbin IS NULL THEN NULL -- Deal with level 0 boundary issue.
+    WHEN quadbin = 5192650370358181887 THEN TRY_TO_GEOGRAPHY(
```
jgoizueta, line 10: We could use TO_GEOGRAPHY here in this cases with fixed input, so we make sure our geometry literals are OK.

modules/quadbin/snowflake/sql/QUADBIN_POLYFILL.sql
```diff
@@ -0,0 +1,53 @@
...
+-- Copyright (C) 2022 CARTO
+----------------------------
+
+CREATE OR REPLACE FUNCTION QUADBIN_POLYFILL
+(geog GEOGRAPHY, resolution INT)
+RETURNS ARRAY
+AS $$
+    IFF(geog IS NULL OR resolution IS NULL,
+        NULL,
+        IFF(resolution < 0 OR resolution > 29,
```
jgoizueta, line 11: 29 -> 26

modules/quadbin/snowflake/sql/QUADBIN_SIBLING.sql
```diff
@@ -0,0 +1,21 @@
...
+        WHEN direction = 'right' THEN
+            _QUADBIN_SIBLING(quadbin, 1, 0)
+        WHEN direction = 'up' THEN
+            _QUADBIN_SIBLING(quadbin, 0, -1)
+        WHEN direction = 'down' THEN
+            _QUADBIN_SIBLING(quadbin, 0, 1)
+        ELSE 
+            NULL
+    END
+$$;
```
jgoizueta, line 21: You could integrate the auxiliar functions into the main one as:

```sql
CREATE OR REPLACE FUNCTION QUADBIN_SIBLING
(quadbin INT, direction STRING)
RETURNS INT
AS $$
  (WITH _dxy AS (
    SELECT CASE direction
        WHEN 'left' THEN
            OBJECT_CONSTRUCT('x', -1, 'y', 0)
        WHEN 'right' THEN
            OBJECT_CONSTRUCT('x', 1, 'y', 0)
        WHEN 'up' THEN
            OBJECT_CONSTRUCT('x', 0, 'y', -1)
        WHEN 'down' THEN
            OBJECT_CONSTRUCT('x', 0, 'y', 1)
        ELSE 
            NULL
    END AS d, QUADBIN_TOZXY(quadbin) AS t    
  )
  SELECT IFF(t:y + d:y >= 0 AND t:y + d:y < BITSHIFTLEFT(1, t:z),
        QUADBIN_FROMZXY(
            t:z,
            MOD(t:x + d:x, BITSHIFTLEFT(1, t:z)) + IFF(t:x + d:x < 0, BITSHIFTLEFT(1, t:z), 0),
            t:y + d:y
        ),
        NULL
    )
   FROM _dxy)
$$;
```

I wrote that function with the hope to avoid the dreaded crashes when using it on a table but it still crashes... I'm afraid we'd need to inline the calls to QUADBIN_TOZXY and QUADBIN_FROMZXY as well, which is nuts :(

modules/quadbin/snowflake/sql/QUADBIN_POLYFILL.sql
vdelacruzb: Also it looks like when we are out of resolutions we rather throw an error. And NULL is more for NULL inputs.

modules/quadbin/snowflake/quadbin.js
volaya: i forgot to remove it, it doesn't make sense in this module, I just got it from the BQ code

modules/quadbin/snowflake/Makefile
```diff
@@ -0,0 +1,4 @@
+MODULE = quadbin
+MODULE_DEPS = quadkey
```
vdelacruzb, line 2: I'm not sure that if need this `MODULE_DEPS`. Normally we only add them in advanced not in core. I see that you are using `QUADINT_TOZXY` but normally the script modulesort.js will take care of sorting the deployment of all the modules.

This kind of topics I think will be nice to polish in Q3 so everything is detected and deployed automatically.

modules/quadbin/snowflake/doc/QUADBIN_KRING.md
```diff
@@ -0,0 +1,33 @@
...
+**Return type**
+
+`ARRAY`
+
+{{% customSelector %}}
+**Example**
+{{%/ customSelector %}}
+
+```sql
+SELECT carto.QUADBIN_KRING(4388, 1);
```
vdelacruzb, line 23: Here the input KRING is wrong.

modules/quadbin/snowflake/test/integration/QUADBIN_ISVALID.test.js
```diff
@@ -0,0 +1,24 @@
...
+});
+
+test('QUADBIN_ISVALID should detect invalid indexes', async () => {
+    const query = 'SELECT QUADBIN_ISVALID(1234) AS OUTPUT';
+    const rows = await runQuery(query);
+    expect(rows.length).toEqual(1);
+    expect(rows[0].OUTPUT).toEqual(false);
+});
+
+/*
```
vdelacruzb, line 17: Why we don't include this test? If not included then the comment can be removed.

Jesus89: Function fixed

Jesus89 (approved):
LGTM!

Let's review in the future the implementation of these functions, to make them more efficient:
- QUADBIN_BOUNDARY
- QUADBIN_KRING
- QUADBIN_KRING_DISTANCES
- QUADBIN_SIBLING

PR #261, opened 2022-06-30
Fix root tile quadbin value

jgoizueta:
Initially I used trailing 0s by mistake, then forgot to fix that value


vdelacruzb (approved):
LGTM

PR #262, opened 2022-07-04
SQL version of Quadbin UDFs in Redshift

vdelacruzb:
Improve Redshift Spatial Indexes Tiler speed [sc-242081]


jgoizueta (approved):

jgoizueta:
Regarding

```
select (4611686018427387904 << 52);
-- 0
select ((4611686018427387904 | (4611686018427387904 << 1)) >> 12);
-- -1125899906842624
```

The first (0) looks right to me, and the problem in the second must be having set the most significative bit to 1 (by the << 1) so that the number becomes negative. My guess is that if we avoid ever setting that bit we should be OK (which is doable for our quadbins)

vdelacruzb:
Ey @jgoizueta @Jesus89 , I've detected that the sql version of quadbin_fromzxy was causing the tiler to slow down. I got back to the python version and the tiler is fast again.

Jesus89 (approved):
LGTM.

Tested performance improvement in QUADBIN_TOZXY.

PR #263, opened 2022-07-05
Fix H3_POLYFILL for large polygons

volaya:
Fix handle large polygon (+180) in H3_POLYFILL for BigQuery and Snowflake.

image


modules/h3/bigquery/sql/H3_POLYFILL.sql
```diff
@@ -14,40 +14,71 @@ AS """
...
         case 'Polygon':
-            polygonCoordinates = [featureGeometry.coordinates];
+            var clippedGeometryA = h3Lib.bboxClip(featureGeometry, bboxA).geometry;
+            polygonCoordinatesA = [clippedGeometryA.coordinates];
+            var clippedGeometryB = h3Lib.bboxClip(featureGeometry, bboxB).geometry;
+            polygonCoordinatesB = [clippedGeometryB.coordinates];
         break;
         default:
-            return null;
+            var bboxGeom = h3Lib.bboxPolygon(h3Lib.bbox(featureGeometry)).geometry;
```
alex-graciano, line 62: Are we ok with the `LINESTRING` case? Is the BBOX used axis aligned? object-aligned?

alex-graciano (changes requested):
Just requested a minor clarification but LGTM

Jesus89 (commented):
Pending:

- Update common/bigquery/version
- Update modules/h3/bigquery/CHANGELOG.md
- Update modules/h3/bigquery/package.json
- Update common/snowflake/version
- Update modules/h3/snowflake/CHANGELOG.md
- Update modules/h3/snowflake/package.json

modules/h3/bigquery/sql/H3_POLYFILL.sql
volaya: The idea here is that all geometries not previously supported are converted now to the their bbox, and then treated as a polygon. Is that what you meant?

alex-graciano: Not exactly. Let's see we have a horizontal line, both AABB and OABB are the same in those cases, but if the line is in a diagonal manner, the OABB is way greater, so the polyfill will be extended not only through the line but in the whole BBOX.

Jesus89: Testing the h3-js polyfill function https://h3geo.org/docs/api/regions/#polyfill, I found it does not work with points or lines.

I have a workaround in mind, which is converting the points and lines into polygons, no bbox but something like a buffer, and then applying polyfill in the buffered section. Buffer distance depends on the resolution: https://h3geo.org/docs/core-library/restable/.

However, I would tackle that in another PR (will create a ticket for the backlog) and revert the bbox changes in this one to keep only the large polygons fix and the max resolution fix.

Jesus89 (changes requested):
Good job!

I have reverted support for points and lines using bbox. We should cover this in a different PR and use a buffer with a distance depending on the H3 resolution.

The only missing part is a couple of tests to cover the two fixes:
- max resolution (0-15)
- polygons larger than 180º

Jesus89 (approved):
LGTM!

PR #270, opened 2022-07-13
Add auxiliar methods for PG tests

jgoizueta:
- [x] New method to upload test PG tables with no geometry
- [x] New method to compare dicts with numeric values approximately

The `dicts_approx_equal` compares dicts and makes comparisons of floats and strings that contain comma-separated lists of numbers approximate (recursively).

### DIFF
--- a/common/postgres/test_utils/__init__.py
+++ b/common/postgres/test_utils/__init__.py
@@ -71,7 +71,7 @@ def floats_approx_equal(a, b, rel_tol=1e-5, abs_tol=0.0):
     return abs(a-b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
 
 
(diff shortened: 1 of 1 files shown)


Jesus89 (approved):

common/postgres/test_utils/__init__.py
```diff
@@ -57,3 +58,60 @@ def upload_csv_to_postgres(table):
...
+
+
+numbers = re.compile(r'^(?:([-+]?\d*\.?\d+)(?:[eE]([-+]?\d+))?\,?)+$')
+
+
+def split_numbers(n):
+    return [float(x) for x in n.split(',')]
+
+
+def floats_list_approx_equal(alist, blist, rel_tol=1e-5, abs_tol=0.0):
```
jgoizueta, line 81: there's a bug here

PR #271, opened 2022-07-14
Return to stable

vdelacruzb:
After some testing, I've seen that the problem was caused by the volatility set to IMMUTABLE. So this is not an option anymore for our Redshfit UDFs.

I've applied the change and also checked that the tiler is pretty fast still. It seems to be barely affected or not at all.


modules/quadbin/redshift/sql/QUADBIN_FROMZXY.sql
```diff
@@ -3,7 +3,7 @@
 ----------------------------
 
 CREATE OR REPLACE FUNCTION @@RS_PREFIX@@carto.QUADBIN_FROMZXY
-(z INT, x INT, y INT)
+(z BIGINT, x BIGINT, y BIGINT)
```
Jesus89, line 6: I would keep INT. Is there any reason to update to BIGINT?

Jesus89 (approved):

modules/quadbin/redshift/sql/QUADBIN_FROMZXY.sql
vdelacruzb: As discussed in Slack the UDF QUADBIN_FROMZXY was complaining when we use columns in a table of type BIGINT. The customers would have to explicitly cast to int. This way we avoid those problems.

PR #273, opened 2022-07-18
Fix lack of precision in UDFs returning arrays

alex-graciano:
Found some lack of precision in several functions returning arrays. In order to solve this issue, the functions now return a STRING for two main issues related with Snowflake limitations:
1. Snowflake has a native support of BigInt numbers, however, if the UDF
returns this data type the next Snowflake internal error is raised:
SQL execution internal error: Processing aborted due to error 300010:3321206824
2. If the UDF returns the hex codification of the quadbin to be parsed in a SQL
higher level by using the _QUADBIN_STRING_TOINT UDF a non-correlated query can be produced.

The reason why the CI did not fail previously was because a loss of precision as well.


modules/quadbin/snowflake/sql/QUADBIN_POLYFILL.sql
```diff
@@ -26,13 +33,16 @@ AS $$
     {
         quadbins = quadbinLib.geojsonToQuadbins(pol, {min_zoom: RESOLUTION, max_zoom: RESOLUTION});
     }
-    return quadbins;
+    let stringQuadbins = '[';
```
Jesus89, line 36: array + join can be used instead

modules/quadbin/snowflake/sql/QUADBIN_TOCHILDREN.sql
```diff
@@ -2,34 +2,54 @@
...
+    }
+
+    if (RESOLUTION < 0 || RESOLUTION > 26) {
+        throw new Error('Invalid resolution, it must be between 0 and 26.');
+    }
 
-    const res = Number(RESOLUTION);
     const tile = quadbinLib.quadbinToTile(INDEX);
+
+    if (RESOLUTION < tile.z) {
```
Jesus89, line 31: Why not using `res` here?

modules/quadbin/snowflake/sql/QUADBIN_TOCHILDREN.sql
```diff
@@ -2,34 +2,54 @@
...
+    }
+
+    const res = Number(RESOLUTION);
     const xmin = tile.x << (res - tile.z);
     const xmax = ((tile.x + 1) << (res - tile.z)) - 1;
     const ymin = tile.y << (res - tile.z);
     const ymax = ((tile.y + 1) << (res - tile.z)) - 1;
 
-    const children = [];
+    let children = '[';
```
Jesus89, line 41: Use array + join instead

modules/quadbin/snowflake/sql/_QUADBIN_KRING.sql
```diff
@@ -13,7 +21,7 @@ AS $$
 
     const offset = Number(SIZE);
     const inputTile = quadbinLib.quadbinToTile(INDEX);
-    const krings = [];
+    let krings = '[';
```
Jesus89, line 24: array + join can be used instead

modules/quadbin/snowflake/sql/QUADBIN_KRING.sql
```diff
@@ -2,12 +2,17 @@
 -- Copyright (C) 2022 CARTO
 ----------------------------
 
--- FIXME: slow
+-- The function returns a STRING for two main issues related with Snowflake limitations
```
Jesus89, line 5: I would move this code to the module's readme

alex-graciano: I'm not pretty sure about this because it's an implementation detail, it should not be impact in the user experience either the user should act in a special way in order to execute the functions.

alex-graciano: I'm not pretty sure about this because it's an implementation detail, it should not be impact in the user experience either the user should act in a special way in order to execute the functions.

PR #277, opened 2022-07-27
Chore/sc 246038/add at databricks scaffolding

aarroyosal:
shortcut ticket - https://app.shortcut.com/cartoteam/story/246038/add-at-databricks-scaffolding
Acceptance:

- Install the libraries needed for the databricks AT (README)
- [Setup databricks databricks CLI authentication (via token)](https://docs.databricks.com/dev-tools/cli/index.html#set-up-authentication)
- test the make commands
    -  make build|clean|lint|deploy CLOUD=databricks
 - test the deployed function in a databricks notebook `select carto_dev_data.st_x(geom) as latitude from carto_dev_data.points_100k limit 100;`


vdelacruzb:
ey @aarroyosal I know that this is a work in progress but I'm going to start adding some doubts and suggestions

.gitignore
```diff
@@ -17,4 +18,10 @@ dev_env.sh
...
 *.ipynb
-.pgpass
\ No newline at end of file
+.pgpass
+
+.idea/
+target/
+.bsp/
+create-job.json
+submit-run.json
```
vdelacruzb, line 27: are this kind of objects created all around the project in different locations os just in a specific folder? Maybe it can be contained in another .gitignore inside the folder.

Makefile
```diff
@@ -1,6 +1,6 @@
 export GIT_DIFF ?= off
 
-.SILENT:
+#.SILENT:
```
vdelacruzb, line 3: Don't forget to get this back before merging :)

README.md
```diff
@@ -162,6 +163,36 @@ AWS_ACCESS_KEY_ID=your-access-key
...
+CARTO Analytics Toolbox for Databricks provides geospatial functionality leveraging the GeoMesa SparkSQL capabilities. It implements Spatial Hive UDFs. In order to install the toolbox the library (jar-with-dependencies) needs to be installed in the cluster you are using, and the Hive UDFs registered via createUDFs sql script
+
+**Tools**
+
+Make sure you have installed the following tools:
+
+- `make`: https://www.gnu.org/software/make/
+- `jdk (8 or 11)`: https://www.oracle.com/java/technologies/javase/javase8u211-later-archive-downloads.html (v8.x)
+- `sbt`: https://www.scala-sbt.org/1.x/docs/Setup.html (v1.x)
+- `Python2.7.9 and above`: https://www.python.org/downloads/release/python-2718/ (v2.7.18)
```
vdelacruzb, line 177: Is python 2 really necessary?

Makefile
```diff
@@ -13,6 +13,10 @@ lint lint-fix install build test-unit test-integration deploy clean clean-deploy
 			echo "> Module $${module}/$(CLOUD)"; \
 			$(MAKE) -C modules/$${module}/$(CLOUD) $@ || exit 1; \
 		done; \
+	elif [ "$(CLOUD)" = "databricks" ]; then \
+		`node scripts/modulesort.js`; \
+		echo "> going with databricks ${ROOT_DIR}"; \
```
vdelacruzb, line 18: This comment could be more general and you can pass the parameter $(CLOUD) in case we reuse the structure for another clouds

clouds/databricks/README.md
```diff
@@ -0,0 +1,161 @@
...
+## Spatial Query Optimizations
+
+There are two types of supported optimizations: `ST_Intersects` and `ST_Contains`, which allow Spark to push down predicates when possible.
+
+To enable optimizations:
+
+```scala
+com.carto.analyticstoolbox.spark.sql.rules.SpatialFilterPushdownRules
+
+val spark: SparkSession = ???
```
vdelacruzb, line 44: This part of the doc has to be completed? or the user has to include his config?

clouds/databricks/core/src/main/resources/sql/createUDFs.sql
```diff
@@ -0,0 +1,90 @@
+CREATE OR REPLACE FUNCTION st_antimeridianSafeGeom as 'com.carto.analyticstoolbox.core.ST_AntimeridianSafeGeom';
```
vdelacruzb, line 1: These are the functions that we called core right? Is it something that we will might be modifying in the future adding more? or they are provided by a third party? In the first case maybe we can generate this sql automatically be readying the other sqls. I wouldn't consider it a priority though.

common/databricks/Makefile
```diff
@@ -0,0 +1,101 @@
...
+	echo "Databricks schema not defined, using default"
+	$(eval DB_SCHEMA := default)
+	echo $(DB_SCHEMA)
+endif
+	sed -i '.bkp' '1s/^/USE ${DB_SCHEMA};\n/' $(SQL_PATH)
+	databricks workspace mkdirs $(SQL_DEPLOY_PATH)
+	databricks workspace import --overwrite --language SQL $(SQL_PATH) $(SQL_DEPLOY_PATH)/createUDFs
+
+create-udfs:
+	sed -e 's/@@DB_CLUSTER_ID@@/${DB_CLUSTER_ID}/g' -e 's!@@SQLPath@@!$(SQL_DEPLOY_PATH)/createUDFs!g' $(ROOT_DIR)/submit-run-template.json > $(ROOT_DIR)/submit-run.json
```
vdelacruzb, line 82: I don't how possible it is and what output we have when we deploy in databricks. It would be really nice to know not only if the job finished ok as you comment below but also which error and which function.

README.md
aarroyosal: You need either python2 or python 3, but now I think that it is better to remove it and assume that everybody will install the databricks CLI with python 3. 
Thanks!

clouds/databricks/.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
+name: CI
```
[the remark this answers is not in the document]
aarroyosal: I have created this story in order to analyze the cost and impact of merging the databricks ci with the general ci 
https://app.shortcut.com/cartoteam/story/247825/analysis-ci-cd-in-at-databricks

clouds/databricks/README.md
aarroyosal: This is when the user creates a spark session. It depends on where is the spark session being created.
For instance the spark session in the databricks notebooks is created automatically and you can access it with the variable spark. (no need for this line in that case)

clouds/databricks/core/src/main/resources/sql/createUDFs.sql
aarroyosal: The package is called core at the moment. For now, they have not been migrated yet, so I just left the sql, but not the UDF that they are calling. 
When we migrate the modules they will have a package like this:
`CREATE OR REPLACE FUNCTION dummy_function as 'com.carto.analyticstoolbox.modules.dummy.DummyFunction';`
The Idea is having each sql script within its own module, although considering that a full package is always going to be delivered, it could be also a good idea to have it centralized in one global script (as is now)

common/databricks/Makefile
aarroyosal: If it fails it is not likely to fail only for certain functions since this queries do not check if the libraries are there, only link the udf to the class. If the class does not exist you will get an error in runtime 
However for errors I will try (if possible) to get an understandable message (cluster not exist, authentication error, not authorized, etc)

.gitignore
aarroyosal: Good idea :)

Makefile
aarroyosal: Good catch

Makefile
aarroyosal: That's right, good catch. Actually I didn't even need to call the modulesort script

Jesus89 (commented):
Good job!

We have commented on some changes for the continuation of this work in the branch `feature/sc-246044/migrate-at-core-databricks-modules`. To avoid duplicating the suggestions, let's open a new draft with that branch and close this one in favor of the new one, so we can do one final review there.

Jesus89:
Superseded by https://github.com/CartoDB/analytics-toolbox-core/pull/288

PR #278, opened 2022-07-27
Concatenate headers instead of overwrite

vdelacruzb:
I am doing this just in BQ for the moment. The think is that the LDS header was overwritten by nothing.


alex-graciano (approved):
LGTM

PR #285, opened 2022-08-09
Apply make_valid to H3 boundaries

Jesus89:
Some H3 boundaries return `null` because the output polygon has some problems:
```
ST_GeogFromText failed: Invalid polygon loop: Edge 0 crosses edge 2
```

Passing `make_valid => TRUE` fixes the polygon before returning the geography.


alex-graciano (approved):
LGTM

PR #296, opened 2022-08-25
H3_COVERINGCELLIDS and S2_CENTER

francois-baptiste:
add S2_CENTER
add H3_COVERINGCELLIDS


Jesus89:
Migrated the code to the new structure and "hide" _H3_COVERINGCELLIDS function, so we can use it, but keep it in an experimental state until we decide to make it public.

Jesus89 (approved):

Jesus89 (approved):

PR #299, opened 2022-08-26
Refactor cloud-based Postgres

Jesus89:
Implement a new structure and migrate all the modules for Postgres. Added CI/CD too.


clouds/postgres/modules/quadbin/sql/__CARTO_ERROR.sql
```diff
@@ -0,0 +1,15 @@
+----------------------------
+-- Copyright (C) 2022 CARTO
+----------------------------
+
+CREATE OR REPLACE FUNCTION @@PG_SCHEMA@@.__CARTO_ERROR(
```
vdelacruzb, line 5: Not needed now but these kind of functions can be moved to a utils module.

Jesus89: Totally right, I would keep if for a future task

vdelacruzb (approved):
LGTM!

PR #305, opened 2022-08-31
Chore/sc 247825/add ci cd in at databricks

aarroyosal:
# Description

- Story: https://app.shortcut.com/cartoteam/story/247825/add-ci-cd-in-at-databricks

Please include a summary of the change. If required, also include relevant info about motivation and context... And if PR includes UI elements (or if it helps for understanding it), please add extra files: snapshots, videos...

## Type of change

- Feature
Add CI/CD jobs to databricks warehouse. 
It is prepared to publish the databricks artifacts in sonatype nexus repository and get the version based on the tags of the repo. Since there are no tags at the moment at the repo, the version of the jars generated will be 0.0.0

# Acceptance

- Push some commit to the repo.
- See the workflow at https://github.com/CartoDB/analytics-toolbox-core/actions/workflows/databricks.yml


.github/workflows/databricks.yml
```diff
@@ -0,0 +1,121 @@
+name: Databricks CI/CD
+
+on:
+  push:
+    paths:
+      - "clouds/databricks"
```
Jesus89, line 6: "clouds/databricks/**"

Jesus89 (commented):
LGTM.

Just a small but important change in the CI script.

I suggest removing also this print  `print(f"executing query '{query}'")` from the tests and use pytest --verbose instead.

.github/workflows/databricks.yml
aarroyosal: Done

Jesus89 (approved):

PR #306, opened 2022-09-01
add appox equal for floats in Redshift tests

ilbambino:


ilbambino:
not needed, closing

Jesus89:
Using https://docs.pytest.org/en/7.1.x/reference/reference.html#id8 instead

PR #307, opened 2022-09-02
Check env variables with regex

Jesus89:
Review env vars documentation. Add regex checks for some: host, bucket, account.


alex-graciano (approved):
LGTM. Tested some success/fail deploy cases both in Redshift and Snowflake and everything worked as expected.

PR #310, opened 2022-09-06
Update makefiles for tiler

alex-graciano:


vdelacruzb (approved):
LGTM. Let's make it in the other branches and then we give it another look if we can polish it and not make that explicit inside core that we are deploying the tiler

clouds/bigquery/common/Makefile
```diff
@@ -5,8 +5,10 @@ NODE_MODULES_DEV = $(COMMON_DIR)/node_modules
 GSUTIL = gsutil
 BQ = bq --location=$(BQ_REGION) --project_id=$(BQ_PROJECT)
 BQ_DEPLOY_DATASET = $(BQ_PREFIX)carto
+BQ_LIBRARY_NAME ?= coreLib
```
vdelacruzb, line 8: this line can be removed, then LGTM

PR #316, opened 2022-09-14
Add linter sql and markdown

vdelacruzb:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/250230/add-linter-sql-and-markdown
- Autolink: [sc-250230]

Add linters for sql and md files.

Some changes to take into account:
- Adding sqlfluff support has been added through a python script to ease the control over the text to lint. Sqlfluff complains when finding `@` so the python code helped to these replacements without intermediate temp files and also to control in which scripts it's failing. I had to add support to python in common for some providers as SF and BQ because of this.
- A space is inserted between markdowns headers and body.
- Added jest.config.js it was required for BQ and SF tests.
- Tox.ini was also required to skip some issues with max_length but it was enough with adding some exceptions to the flake8 command to make it work.
- Variables are forced to lower everything else to upper. Existing camelcase variables and arguments are being replaced by snake case.
- In case the sql linter is stuck it exits 1 and return the failed file. Sometimes is not trivial to know why it's stuck and it will require some manual fixes before everything runs ok, normally it could be an indentation issue.
- Everyone should use tab size equal to 4 spaces. Many scripts have been found with 2 spaces, some with 6 and another with 8.

Missing linting on pre-commit. Right now linting could take a while for some modules when we add that feature let's make sure only diff files are linted.

## Type of change

- Feature

# Acceptance

1. Go to any SQL file not added to sqlfluffignore at any provider. Replace a keywork with lower case, example 'CREATE TABLE' by 'create table'.
2. Use make lint.
3. See how it gets back to 'CREATE TABLE'.
4. Remove a space between a header and body at any markdown file.
5.  Use make lint.
6. See how the space is back.


clouds/bigquery/README.md
```diff
@@ -6,15 +6,16 @@ The Analytics Toolbox for Bigquery contains SQL functions and JavaScript librari
 
 Make sure you have installed the following tools:
 
-- `make`: https://www.gnu.org/software/make/
-- `node`: https://www.npmjs.com/ (v14.18)
-- `yarn`: https://yarnpkg.com/ (v1.22)
-- `bq`: https://cloud.google.com/bigquery/docs/bq-command-line-tool
-- `gsutil`: https://cloud.google.com/storage/docs/gsutil (v5.5)
+- `make`: <https://www.gnu.org/software/make/>
```
Jesus89, line 9: Add Python3

Jesus89: _- `Python3`: <https://www.python.org/downloads/release/python-3811> (v3.8.11)_

Jesus89 (approved):

PR #329, opened 2022-09-28
Use PG_SCHEMA to load/drop tables in test utils

Jesus89:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/260569
- Autolink: [sc-260569]

Small PR to use PG_SCHEMA in the loading and dropping of tables used in the tests. That way we will avoid possible collisions by using the default schema.

## Type of change

- Fix


vdelacruzb (approved):
LGTM

PR #330, opened 2022-09-29
make QUADBIN_TOCHILDREN deterministic

francois-baptiste:
# Description

Add order with clause to make children order deterministic in QUADBIN_TOCHILDREN output array.


Jesus89:
This makes the function slower. We need to discuss why is this required, and what's the cost of it

francois-baptiste:
I was needing this for unpacking quadbin raster.
Accepting this PR would have avoided code duplication.

Jesus89:
OK, we'll take it into account in the code review of the raster module

PR #338, opened 2022-10-14
add S2_RESOLUTION and S2_TOCHILDREN

francois-baptiste:
# Description

add function S2_RESOLUTION
add function S2_TOCHILDREN

PR #348, opened 2022-10-28
Research point linestring support in h3 polyfill

vdelacruzb:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/243353/research-point-linestring-support-in-h3-polyfill

Allow H3_POLYFILL to work with lines and points. The approach followed consist of getting the S2 boundaries that covers a geography and intersect the H3 taken from their centroids with the original geography. This is not 100% precise as depends on the S2 size taken. I've added a function to approximate values that work but we will always find situation where certain edges could be omitted. If we use a S2 level too hight the query would take too long so it's a trade-off between accuracy and speed.

## Type of change

- Refactor

# Acceptance

I've adjusted the integration tests as they now allow POINTS and LINESTRINGS and done some test with points and lines.

Also have check some visualizations to see that the H3 really intersects the origin geography.

Captura de pantalla 2022-11-03 a las 17 34 54


vdelacruzb:
This PR depends on https://github.com/CartoDB/analytics-toolbox-core/pull/338.

vdelacruzb:
I've restored the Snowflake back to normal. I have been facing issues with the new approach when modifying the integration tests. Having the new approach would cause `H3_POLYFILL` not working with CTEs or any dynamic value as we can see in the next code.

```
WITH inputs AS
(
  SELECT 11 AS id, TO_GEOGRAPHY('POINT(0 0)') as geom, 15 as resolution UNION ALL
  SELECT 12 AS id, TO_GEOGRAPHY('MULTIPOINT(0 0, 1 1)') as geom, 15 as resolution
  SELECT 13 AS id, TO_GEOGRAPHY('LINESTRING(0 0, 1 1)') as geom, 3 as resolution UNION ALL
  SELECT 14 AS id, TO_GEOGRAPHY('MULTILINESTRING((0 0, 1 1), (2 2, 3 3))') as geom, 3 as resolution UNION ALL

  -- 15 is a geometry collection containing only not supported types
  SELECT 15 AS id, TO_GEOGRAPHY('GEOMETRYCOLLECTION(POINT(0 0), LINESTRING(1 2, 2 1))') as geom, 1 as resolution
)
 SELECT CARTO_BACKEND_DATA_TEAM.vdelacruz_carto._H3_LINES_POLYFILL(geom, RESOLUTION)
from inputs;
-- SQL compilation error: Unsupported subquery type cannot be evaluated
```

We have faced this issue in other situations and it's a Snowflake limitation cause by correlated subqueries. We have it as feature request and more people are asking for the same: https://community.snowflake.com/s/question/0D50Z00008uicMqSAI/unsupported-subquery-type-cannot-be-evaluated-in-udft-when-using-or-in-where-clause

This same issue is present in others UDFs as DELAUNAYLINES for instance. In that case the test were created avoiding CTEs, but I wouldn't recommend do the same to a Polyfill function that it's used in plenty of cases.

francois-baptiste:
Nice work @vdelacruzb, this way we get best of both world in one function.

I got one thing disturbing however, some cells are missing in this linestring polyfill:

```sql
SELECT *
FROM UNNEST(`cartodb-gcp-backend-data-team.fbaptiste_carto.H3_POLYFILL`(ST_MAKELINE(ST_GEOGPOINT(0,0),ST_GEOGPOINT(75,75)), 10))
-- original linestring with tesselation SELECT ST_GEOGFROMGEOJSON(ST_ASGEOJSON(ST_MAKELINE(ST_GEOGPOINT(0,0),ST_GEOGPOINT(75,75))))
```

image

vdelacruzb:
Ey @francois-baptiste I'm afraid that this is an issue that we might have in some cases. The only thing that we can do is increase the s2 generation resolution for a given H3 resolution. I've just increased the resolution and as you can see it seems ok now.

Captura de pantalla 2022-11-04 a las 10 35 48

As I commented above it's kind of a trade-off between speed and accuracy. I have tried to increase the rest of the S2 levels a bit more but for instance I already get timeouts in the integration test for a h3 resolutions of 1 and 3 which is not that high.

Another test that I have tried is polyfilling circles with this method and I see how even if the circle looks good when I increase the S2 resolution there are few H3 at the edge that are added.

francois-baptiste:
> Ey @francois-baptiste I'm afraid that this is an issue that we might have in some cases. The only thing that we can do is increase the s2 generation resolution for a given H3 resolution. I've just increased the resolution and as you can see it seems ok now.
> 
> Captura de pantalla 2022-11-04 a las 10 35 48
> 
> As I commented above it's kind of a trade-off between speed and accuracy. I have tried to increase the rest of the S2 levels a bit more but for instance I already get timeouts in the integration test for a h3 resolutions of 1 and 3 which is not that high.
> 
> Another test that I have tried is polyfilling circles with this method and I see how even if the circle looks good when I increase the S2 resolution there are few H3 at the edge that are added.

Ok, thanks for the explanation @vdelacruzb . This tradeoff seems ok for me for the time being.

clouds/bigquery/modules/doc/h3/H3_POLYFILL.md
```diff
@@ -6,9 +6,9 @@ carto.H3_POLYFILL(geography, resolution)
 
 **Description**
 
-Returns an array with all the H3 cell indexes **with centers** contained in a given polygon. It will return `null` on error (invalid geography type or resolution out of bounds).
+Returns an array with all the H3 cell indexes **with centers** contained in a given polygon. It will return `null` on error (invalid geography type or resolution out of bounds). In case of lines, it will return the H3 cell indexes intersecting those lines. For a given point, it will return the H3 index of cell in which that point is contained.
```
ilbambino, line 9: I think we should add a comment regarding the known issues, that in some cases some cells might be missing due to the approximations we do

ilbambino (approved):
except that I would add some comment to the doc 🚀

ilbambino (approved):

PR #350, opened 2022-11-16
Chore: create github workflows for the release process 

aarroyosal:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/266722/create-a-gh-actions-workflow-to-start-at-atc-releases
- Autolink: [sc-266722]

Create workflow to draft a new release based on this document https://docs.google.com/document/d/1VQPp4Au3D_LMGRCiiX6HN0bIg4RZI-1Sb3be7j-XYCg/edit

## Type of change

- Chore: create github workflows for the release process 

# Acceptance

Go to github actions, workflow draft-new-release and run a new job passing the version name (2022-11)

# Basic checklist

- Good PR name
- Shortcut link
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation


.github/workflows/bigquery.yml
```diff
@@ -90,10 +91,10 @@ jobs:
       - name: Run deploy
         run: |
           cd clouds/bigquery
-          make deploy diff="$GIT_DIFF" production=1
+          make deploy diff="$GIT_DIFF"
```
vdelacruzb, line 94: Removing production=1 from here might be the reason why you were getting a versión including `-dev`.

aarroyosal: No, because I was getting it in the test, that is before this step. 
The reason was that it was doing a substring of the first 10 characters, to get the version in format yyyy-MM-dd, but in the new format x.y.z that was not valid. I think that's already fixed in the tests

.github/workflows/bigquery-ded.yml
```diff
@@ -18,13 +18,29 @@ jobs:
...
     env:
       GCLOUD_VERSION: 290.0.1
-      BQ_PREFIX: dedicated_${{ github.event.pull_request.number }}_
       BQ_PROJECT: ${{ secrets.BQ_PROJECT_CD }}
       BQ_REGION: ${{ secrets.BQ_REGION_CD }}
       BQ_BUCKET: ${{ secrets.BQ_BUCKET_CD }}
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
+      - name: Set BQ_PREFIX for releases
```
vdelacruzb, line 27: I'm not sure if we want this for the dedicateds. Maybe this was defined as it is. Normally a dedicated consist of deploying temporally our UDFs in a controlled env so we can test it before merging and then just remove that temp env. Even if we have an approach for hot fixes I believe we will need dedicateds still for QA

aarroyosal: I am not sure I understand the problem. This was not defined, but I just thought it would be more descriptive to deploy releases and hotfixes with different names like 'dedicated_release_version_' instead of 'dedicated_prNumber_' like the regular dedicated schemas.
The regular dedicated will continue working exactly like they were, or that's the idea at least

vdelacruzb: I see what you mean. I'm worried then if two of us are working in different PRs and want just to do some tests and update the version once everything is ready we would have a conflict I guess. PR_Number was a way to have totally independent dedicateds. If you think there is no possibility of conflicts then it LGTM.

vdelacruzb (approved):
LGTM

PR #371, opened 2022-12-20
fix: H3_BOUNDARY in Snowflake by removing duplicated points

### LINKED ISSUE #351: carto.H3_BOUNDARY returns null for some valid cell IDs
Hi! Thank you so much for the toolbox, it really makes my life easier! :)

Now, on to the bug:

The `carto.H3_BOUNDARY` function returns null for some valid cell IDs.

1. Use Snowflake
2. Run these three queries:
    a. `SELECT carto.carto.H3_BOUNDARY('870813269ffffff');`
    b. `SELECT carto.carto.H3_BOUNDARY('8708132c9ffffff');`
    c. `SELECT carto.carto.H3_BOUNDARY('870811a69ffffff');`
3. All of these queries return `NULL`.

I was expecting all the queries above to return a GEOGRAPHy representing the H3 cell with the corresponding ID. I have inspected the cells with these IDs [here](https://wolf-h3-viewer.glitch.me/), and they seem perfectly fine.

Thanks!

Jesus89:
# Description

Closes #351

## Type of change

- Fix
- Tests

# Acceptance

```sql
SELECT carto.carto.H3_BOUNDARY('870813269ffffff');
-- POLYGON((... instead of NULL
```


clouds/snowflake/libraries/javascript/libs/h3_boundary.js
```diff
@@ -1,6 +1,27 @@
 import { h3IsValid, h3ToGeoBoundary } from '../src/h3/h3_boundary/h3core_custom';
 
+function removeDuplicates (coordinates) {
+    const precision = 0.0000000000001;
+    const uniqueCoordinates = new Set();
+
+    for (const coordinate of coordinates) {
+        let isUnique = true;
+        for (const uniqueCoordinate of uniqueCoordinates) {
```
jgoizueta, line 9: We could compare only with the previous vertex (modulo the number of vertex) and safe a few comparisons, not a big deal for such small polygons, but for 6 coordinates we'd do 12 coordinate comparisons  instead of 30

Jesus89: Good point

jgoizueta (approved):

PR #381, opened 2023-01-20
fix(bq,sf): use extended toBeCloseTo in tests

Jesus89:
Extend `toBeCloseTo` in BigQuery and Snowflake tests for objects


vdelacruzb (approved):
LGTM. It's nice to have this.
I guess we have this in a separate file and not inside `common/test-utils.js` to avoid clients instantiation from the libraries tests.

We could potentially update that in the future:
- cloud/common/test-utils:
  - client.js
  - extend.js
  - fixtures.js
  - sort_norm.js
  - (...)

PR #406, opened 2023-05-25
fix(tools|installer): verify lds config before installing

vdelacruzb:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/311402/add-params-validator-in-the-at-installer-script
- Autolink: [sc-311402]

## Type of change

- Fix
- 
# Acceptance

Tested by removing LDS config values, modifying each of them and removing fields from the jwt.

### DIFF
--- a/tools/installer/src/__init__.py
+++ b/tools/installer/src/__init__.py
@@ -5,6 +5,7 @@ import yaml
 import click
 import zipfile
 import redshift_connector
+import jwt
 
 from tqdm import trange
 from sqlparse import split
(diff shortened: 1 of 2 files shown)


tools/installer/src/__init__.py
```diff
@@ -77,6 +78,31 @@ def run_sql(sql, config):
...
+def validate_lds_config(lds_config):
+    pattern = r'^(lds-function-asia-northeast1|lds-function-australia-southeast1|lds-function-europe-west1|lds-function-us-east1)$'  # noqa: E501
+    if not validate_str(lds_config.get('lambda'), pattern):
+        exit('incorrect configuration: missing or invalid lds.lambda')
+
+    pattern = r'^arn:aws:iam::[0-9]+:role/CartoFunctionsRedshiftRole,arn:aws:iam::000955892807:role/CartoFunctionsRole$'  # noqa: E501
+    if not validate_str(lds_config.get('roles'), pattern):
+        exit('incorrect configuration: missing or invalid lds.roles')
+
+    if not validate_str(lds_config.get('api_base_url')):
```
luipir, line 90: IMHO you could also validate more building URL(lds_config.get('api_base_url')) and check if it trigger errors

vdelacruzb: Is it ok if I just check if it returns a 200 as status code? The thing is this URL won't be necessary a CARTO URL as It could be self hosted

luipir: I mean create a URL class instalce `new URL(lds_config.get('api_base_url'))` and catch error if it is not valid. e.g. do not test if it is reachable but just it is formally correct URL string

vdelacruzb: ok. I just added a URL validator. It's python so I couldn't just do `new URL`

luipir (approved):

PR #443, opened 2023-08-23
chore(sf): add some auxiliary test-utils functions

vdelacruzb:
# Description

They are the same that we had in BQ. Just added a new function that converts the keys from arrays of dictionaries to lower case. As depending how a select is done keys could be upper or lower case. Maybe this could just be included inside the sorting function, but didn't want to modify it.

Also I run `make lint` that's the reason of the other change


luipir:
just a doubt... why this utilities only for SF?

vdelacruzb:
> just a doubt... why this utilities only for SF?

Actually they are already present in BQ. I just copy pasted them.
They are using in multiple tests in BQ but we were not using those in SF even if they are useful.

PR #445, opened 2023-09-18
fix(bq,sf): avoid linting node_modules

vdelacruzb:
# Description

Shortcut

- Autolink: [sc-342684]

It seems that in some cases node_modules could include python files. This code update should exclude those files.

## Type of change

- Fix


Jesus89 (approved):
Quick :heavy_check_mark: 

But test first with the submodule pointing to the feature branch, and then we can merge.

vdelacruzb:
> Quick ✔️
> 
> But test first with the submodule pointing to the feature branch, and then we can merge.

Yes, I was doing that, in the other branch. Once the tests passes I will proceed.

vdelacruzb:
Update: Tests passed

PR #454, opened 2023-12-15
feat(sf): deploy analytics toolbox native app

vdelacruzb:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/353844/build-analytics-toolbox-native-app-package
- Autolink: [353844]

This PR includes the infrastructure to deploy automatically Native Apps versions. Contains some logic to, depending on which kind of version change we have deploy a whole new version, patch, delete previous versions to not surpass the 2 versions limits. Notice that the uploaded versions are called `v<major>`(v1, v2...) and then we include the whole version in the label so it will be still visible by the customers.

Also will deploy a native app on each merge in order to check that the upgrade keeps doing successfully, as it could crash if anyone use a Temp table in a procedure for instance.

I included a simple README even if this package is not focused on production.

## Type of change

- Feature

# Acceptance

I've run some test regarding deployment, re-deployment, updating versions, seeing which functions are available: https://app.shortcut.com/cartoteam/story/367433/native-apps-installation-qa


jgoizueta (approved):

Jesus89 (approved):
Great job!

PR #483, opened 2024-02-20
chore(rs,pg): remove installer tool

vdelacruzb:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/383859/update-setup-procedures-to-use-at-gateway
- Autolink: [sc-383859]

Looking at the previous state of the script, I think we might be missing `RS_` as well so the replacements can be applied successfully.

## Type of change

- Refactor

## Acceptance

This code still have to be updated and tested when the new endpoints are added.
As we are currently checking:
```
    pattern = r'^(lds-function-asia-northeast1|lds-function-australia-southeast1|lds-function-europe-west1|lds-function-us-east1)$'  # noqa: E501
```
And they should be updated.


Jesus89:
@ernesmb could you confirm if we can proceed with the merge?

ernesmb:
yes, confirmed we can remove it

Jesus89 (approved):

PR #489, opened 2024-03-22
Chore/sc 391760/reimplement basic h3 functions in at snowflake

DeanSherwin:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/STORY_ID
- Autolink: [sc-STORY_ID]

## Type of change

- Refactor

Change our H3 functions to use Snowflakes new native H3 functions. 

# Acceptance

View or run the snowflake_test.sh benchmark test script (available here - https://app.shortcut.com/cartoteam/story/391762/run-benchmarks-with-new-at-snowflake-h3-functions). It runs each of the re-implemented functions on 'main' and this branch and dumps execution times to CSV.


clouds/snowflake/modules/test/h3/H3_TOCHILDREN.test.js
```diff
@@ -71,4 +71,4 @@ test('Coarser resolution returns empty array', async () => {
     const rows = await runQuery(query);
     expect(rows.length).toEqual(1);
     expect(rows[0].TOP_CHILDREN).toEqual([ ]);
-});
\ No newline at end of file
+});
```
vdelacruzb, line 74: I see some spaces around. Please try `make lint` and then submit the results. We have pending adding a pre-commit to ensure this linting but it's a good practice to do it.

clouds/snowflake/modules/sql/h3/H3_CENTER.sql
```diff
@@ -1,31 +1,11 @@
 ----------------------------
--- Copyright (C) 2023 CARTO
+-- Copyright (C) 2021-2024 CARTO
 ----------------------------
```
vdelacruzb, line 3: This is something small but we try to keep the `-` to the end. You can just copy the header from another file and modify it. This would apply to the rest of the files of this PR.

```
----------------------------
-- Copyright (C) 2021-2024 CARTO
----------------------------
```
->
```
--------------------------------
-- Copyright (C) 2021-2024 CARTO
--------------------------------
```

vdelacruzb:
I would update the Description -> Shortcut on this PR to contain the shortcut story and autolink.

Also I recommend taking a look to the [`CONTRIBUTING.md`](https://github.com/CartoDB/analytics-toolbox-core/blob/chore/sc-391760/reimplement-basic-h3-functions-in-at-snowflake/CONTRIBUTING.md#L3). You will find more details about conventions for PR titles.

In this case would be more like: `chore(sf|h3): reimplement basic h3 functions`

clouds/snowflake/modules/sql/h3/H3_BOUNDARY.sql
```diff
@@ -1,32 +1,15 @@
...
-    }
-
-    const coords = h3BoundaryLib.h3ToGeoBoundary(INDEX, true);
-    const uniqueCoords = h3BoundaryLib.removeNextDuplicates(coords);
-    return `POLYGON((${uniqueCoords.map(c => `${c[0]} ${c[1]}`).join(',')}))`;
-$$;
-
 CREATE OR REPLACE SECURE FUNCTION @@SF_SCHEMA@@.H3_BOUNDARY
-(index STRING)
+(h3_hex STRING)
```
vdelacruzb, line 6: Even if h3_hex sounds more clear I wouldn't update it, as we just use the name `index` in many places so it fits more the current implementation.

If at some point we need to modify parameters names remember that it's also necessary to modify them in the corresponding doc. In this case: https://github.com/CartoDB/carto-analytics-toolbox-core/blob/04bbf15c6a0a4c9c8d14ec7e60af056f94d58cc9/clouds/snowflake/modules/doc/h3/H3_BOUNDARY.md

clouds/snowflake/modules/sql/h3/H3_KRING.sql
```diff
@@ -1,30 +1,15 @@
...
-    if (!h3KringLib.h3IsValid(ORIGIN)) {
-        throw new Error('Invalid input origin')
-    }
-
-    return h3KringLib.kRing(ORIGIN, parseInt(SIZE));
-$$;
-
 CREATE OR REPLACE SECURE FUNCTION @@SF_SCHEMA@@.H3_KRING
-(origin STRING, size INT)
+(h3_hex STRING, distance INT)
```
vdelacruzb, line 6: For the distance occurs the same that for `index` previously. Don't know if I would keep size or distance.
What do you think @ernesmb?

In any case if we decide to go for distance this should be updated in other provider too and also in every doc to be consistent.

clouds/snowflake/modules/sql/h3/H3_KRING_DISTANCES.sql
```diff
@@ -1,38 +1,28 @@
 ----------------------------
--- Copyright (C) 2021 CARTO
+-- Copyright (C) 2021-2024 CARTO
 ----------------------------
 
-CREATE OR REPLACE FUNCTION @@SF_SCHEMA@@._H3_KRING_DISTANCES
-(origin STRING, size DOUBLE)
+CREATE OR REPLACE SECURE FUNCTION @@SF_SCHEMA@@._H3_KRING_DISTANCES
+(origin STRING, hexarray ARRAY)
```
vdelacruzb, line 6: Sam about naming.

In this function I don't know why added `SECURE` and also I think it should keep being `IMMUTABLE` as the results should be deterministic.

clouds/snowflake/modules/sql/h3/H3_TOPARENT.sql
```diff
@@ -1,30 +1,15 @@
...
-$$;
-
 CREATE OR REPLACE SECURE FUNCTION @@SF_SCHEMA@@.H3_TOPARENT
-(index STRING, resolution INT)
+(h3_hex STRING, resolution INT)
 RETURNS STRING
 IMMUTABLE
 AS $$
-    @@SF_SCHEMA@@._H3_TOPARENT(INDEX, CAST(RESOLUTION AS DOUBLE))
+    IFF(
```
vdelacruzb, line 10: The same way that we have added a condition to `H3_TOCHILDREN` to see if the parameter resolution is bigger. The same could be applied here but with smaller resolutions.

It might be also ok in both cases to check that the resolution is in [0,15], as done in `H3_FROMLONGLAT`.

clouds/snowflake/modules/sql/h3/H3_KRING.sql
DeanSherwin: For context: My main reason for changing these is that 'index' and 'size' are reserved keywords in SQL. 

As for the specific new terms I chose: 
The 'distance' word is what is used in the official Uber H3 libraries. Example:  https://uber.github.io/h3-py/api_reference#h3.great_circle_distance

And I use either h3_hex or h3_int in any function which takes a H3 index to differentiate between the hex and int representation of H3 cells.

vdelacruzb: To me at least changing size -> distance sounds oki.

Regarding `index`-> `h3_hex`/`h3_int` not that sure.

cc: @Jesus89

clouds/snowflake/modules/test/h3/H3_TOCHILDREN.test.js
DeanSherwin: Done. It turns out I didn't have the `virtualenv` command locally. So my linting was only partial. Fixed now.

clouds/snowflake/modules/sql/h3/H3_COMPACT.sql
```diff
@@ -1,6 +1,6 @@
-----------------------------
+--------------------------------
 -- Copyright (C) 2021 CARTO
-----------------------------
+--------------------------------
```
vdelacruzb, line 3: Not need this change here. So can keep the file as it was.

clouds/snowflake/modules/sql/h3/H3_CENTER.sql
```diff
@@ -1,31 +1,11 @@
...
-    @@SF_LIBRARY_H3_CENTER@@
-
-    if (!h3CenterLib.h3IsValid(INDEX)) {
-        return null;
-    }
-
-    const center = h3CenterLib.h3ToGeo(INDEX);
-    return `POINT(${center[1]} ${center[0]})`;
-$$;
+--------------------------------
```
vdelacruzb, line 1: 2023-2024

clouds/snowflake/modules/sql/h3/H3_INT_TOSTRING.sql
```diff
@@ -1,8 +1,12 @@
+---------------------------------
+-- Copyright (C) 2021-2024 CARTO
```
vdelacruzb, line 2: This was created in 2022.
So 2022-2024

Jesus89:
Hi @DeanSherwin @vdelacruzb.

I think renaming the input parameters is not needed (that way we avoid the need to update the docs).

I've tested with index, INDEX, "INDEX", and it worked fine. What problem did you face @DeanSherwin?

DeanSherwin:
> Hi @DeanSherwin @vdelacruzb.
> 
> I think renaming the input parameters is not needed (that way we avoid the need to update the docs).
> 
> I've tested with index, INDEX, "INDEX", and it worked fine. What problem did you face @DeanSherwin?

Not a technical problem, Just a habit. I like to avoid using reserved SQL keywords like INDEX or SIZE as parameter names. 

I will change it back now. Thanks!

Jesus89:
Got it. Let's keep the interface for now to minimize the changes, but good point to consider in the future

DeanSherwin:
@Jesus89 @vdelacruzb Guys - my thinking now is that H3_KDISTANCE_RING should be reverted to the original UDF as on `main` branch. 

A single call to a JS library is much faster than using the new H3_GRID_DISK and then passing that array to a UDF for processing. My attempts at a better algorithm in JS failed. And any SQL I can create to get a better performance isn't compatible with a UDF. Thoughts?

Unlike H3_BOUNDARY I don't think this warrants a follow-up task as this function doesn't have a native SF implementation and is custom to CARTO's UDF library anyways.

Jesus89:
I agree. We could create a ticket to research in the future (same as H3_BOUNDARY) if H3_GRID_DISK gets improved or they provide a more useful function for our H3_KDISTANCE_RING.

DeanSherwin:
> I agree. We could create a ticket to research in the future (same as H3_BOUNDARY) if H3_GRID_DISK gets improved or they provide a more useful function for our H3_KDISTANCE_RING.

Agreed: https://app.shortcut.com/cartoteam/story/398983/research-an-improved-h3-kring-distance-for-at-sf

clouds/snowflake/results.csv
```diff
@@ -0,0 +1,18 @@
+* SnowSQL * v1.2.31
```
vdelacruzb, line 1: we should remove this file

vdelacruzb (approved):
Nice job. LGTM

Once you have removed this file: clouds/snowflake/results.csv
You can proceed with merging

PR #499, opened 2024-04-29
fixed issue with table names in h3_polyfill_table

volaya:
# Description

Minor fix to correct the WHERE condition used in H3_POLYFILL_TABLE, which was not working instead of the table name being a FQN


DeanSherwin:
Code looks fine. And I tested it manually and it's working. 

Only feedback:

We should add tests for the new `_GET_TABLENAME` function. There are existing tests for `_GET_SCHEMA` and `_GET_DATABASE` at `clouds/snowflake/modules/tests/utils/` which you can copy.

volaya:
Sure, I will add a test. I guess we should also add tests for the H3_POLYFILL_TABLE function. Let me know what you think @DeanSherwin , and I can take care of that as well.

DeanSherwin:
> Sure, I will add a test. I guess we should also add tests for the H3_POLYFILL_TABLE function. Let me know what you think @DeanSherwin , and I can take care of that as well.

More tests sounds good to me!

I think I did not test the _TABLE function as there's no if/else or other logic; it just called H3_POLYFILL. But in hindsight it would be good to test the happy-path (table created, columns match etc..) and also an edge case or two. Eg.the output table already exists.

clouds/snowflake/modules/test/utils/GET_TABLENAME.test.js
```diff
@@ -0,0 +1,22 @@
+const { runQuery } = require('../../../common/test-utils');
+
+describe('_GET_TABLENAME should detect the correct TABLENAME from table identifier', () => {
```
DeanSherwin, line 3: I propose something like:

```
const { runQuery } = require('../../../common/test-utils');

describe.each([
    'databasename.schemaname.tablename',
    'schemaname.tablename',
    'tablename'
])('_GET_TABLENAME should extract the table name from any valid table identifier', (tableIdentifier) => {
    test(`When identifier is ${tableIdentifier}: should return 'tablename'`, async () => {
        const query = `SELECT _GET_TABLENAME('${tableIdentifier}') AS result`;
        expect(rows[0].RESULT).toEqual('tablename');
    });
});
```
Which is less code and yields a slightly more readable output of:
```
Testing modules...
 PASS  test/utils/GET_TABLENAME.test.js
  _GET_TABLENAME should extract the table name from any valid table identifier
    ✓ When identifier is databasename.schemaname.tablename: should return 'tablename' (863 ms)
    ✓ When identifier is schemaname.tablename: should return 'tablename' (168 ms)
    ✓ When identifier is tablename: should return 'tablename' (216 ms)
```

clouds/snowflake/modules/test/h3/H3_POLYFILL_TABLE.test.js
```diff
@@ -0,0 +1,24 @@
+const { runQuery, deleteTable } = require('../../../common/test-utils');
+
+test('H3_POLYFILL_TABLE should work', async () => {
```
DeanSherwin, line 3: I think this test can be improved in one of two ways:

## Option 1: Quick
Just change the name from `H3_POLYFILL_TABLE should work` to `H3_POLYFILL_TABLE should create a new table which stores the result of H3_POLYFILL`

## Option2:  In-depth
1. The name `H3_POLYFILL_TABLE should work` to `H3_POLYFILL_TABLE should create a new table with the output of H3_POLYFILL`

2. Use describe.each or test.each to pass in each mode (intersects, contains, center). Call the H3_POLYFILL_TABLE procedure using the new 'mode' parameter. 
3. Instead of testing against an expected row length, compare it to the output of an inline query to H3_POLYFILL (again, also using the mode parameter).

clouds/snowflake/modules/sql/utils/_CHECK_TABLE.sql
```diff
@@ -0,0 +1,23 @@
+---------------------------
+-- Copyright (C) 2024 CARTO
+---------------------------
+
```
DeanSherwin, line 4: 1. Rename this to `_VALIDATE_IDENTIFIER` or similar. By the current name I would assume it actually checks a real table exists or doesn't exist or something. 

2. It would be good to have a test for this for consistency with the other utils. My suggestion would be two describe.each . One for VALID identifiers. Another for invalid identifiers. But even one simple happy-path test would suffice.

clouds/snowflake/modules/sql/h3/H3_POLYFILL_TABLE.sql
```diff
@@ -15,18 +15,38 @@ EXECUTE AS CALLER
...
+        SELECT LISTAGG(COLUMN_NAME, '','')
+        FROM ' || database_name || '.INFORMATION_SCHEMA.COLUMNS
+        WHERE TABLE_NAME ILIKE @@SF_SCHEMA@@._GET_TABLENAME(''' || output_table || ''')
+        AND TABLE_CATALOG ILIKE @@SF_SCHEMA@@._GET_DATABASE(''' || output_table || ''')
+        AND TABLE_SCHEMA ILIKE @@SF_SCHEMA@@._GET_SCHEMA(''' || output_table || ''')
+        AND COLUMN_NAME NOT ILIKE ''geom''
+        ';
+
+    EXECUTE IMMEDIATE column_names_query;
+    SELECT $1 INTO column_names_csv FROM TABLE(RESULT_SCAN(LAST_QUERY_ID()));
```
DeanSherwin, line 44: Just curious as to why we now execute this separately and retrieve the result. Does this help prevent injection?

DeanSherwin (changes requested):
Most of the changes are to do with tests. I know these are copied from MY tests - but since this original task I've learned more features of Jest which I think we should use. I also think we should abandon the `'X function should work`' naming habit as it doesn't describe or help future devs really. All functions should work :)

Jesus89:
I see the point of getting the database for INFORMATION_SCHEMA.COLUMNS, but after a thought, I would say this is not required because the query can be simplified. I will create an alternative proposal.

Jesus89:
I've added a simple alternative in this PR: https://github.com/CartoDB/analytics-toolbox-core/pull/501

PR #510, opened 2024-05-21
release: 2024-05-21

vdelacruzb:
# Release 2024-05-21

### Changelog
- refactor(sf|h3): avoid memory limit exceeded in H3_POLYFILL_TABLE (#501)
- docs(bq,sf,rs|constructors): fix ST_MAKEELLIPSE angle parameter explanation (#505)
- chore(sf): update create-package and licenses year (#507)
- chore(pg): DB Connection Error in tests when % character in PG_PASSWORD (#509)

### Version
- snowflake: v1.2.3


clouds/postgres/version
```diff
@@ -1 +1 @@
-1.3.0
+1.3.1
```
Jesus89, line 1: Don't need to bump release only with internal chore. Please revert to 1.3.0 and we will keep the change only in the global changelog

clouds/redshift/version
```diff
@@ -1 +1 @@
-1.1.1
+1.1.2
```
Jesus89, line 1: Don't need to bump release only with docs. Please revert to 1.1.1 and we will keep the change only in the global changelog

Jesus89 (changes requested):
Reviewed. We should not bump for docs and internal chores.

Let's update the changes and bump only Snowflake

Jesus89:
:thinking: there are a lot of changes and conflicts, could you check that?

Jesus89 (approved):
LGTM

Remember to update the versions in the description

PR #514, opened 2024-07-02
release: 2024-06-27

Jesus89:
# Release 2024-06-27 (sync with AT release)

### Changelog

- chore(sf): refactor at snowflake native app to an installer (#512)
- chore(bq|quadbin): optimize polyfill (#513)

### Version
- bigquery: v1.2.3
- snowflake: v1.2.4

PR #515, opened 2024-07-02
test: fix h3 bundle in sf

Jesus89:
context https://github.com/CartoDB/analytics-toolbox-core/pull/514#issuecomment-2202637216

## Problem

Recently, without any change from our side, a specific bundle h3_polyfill can not be properly generated:

```
/home/runner/work/analytics-toolbox-core/analytics-toolbox-core/clouds/snowflake/libraries/javascript/libs/h3_polyfill.js → build/h3_polyfill.js...
(!) Missing shims for Node.js built-ins
Creating a browser bundle that depends on "util". You might need to include https://github.com/FredKSchott/rollup-plugin-polyfill-node
Created bundle h3_polyfill.js: 172.17 kB → 50.83 kB (gzip)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
util (guessing 'require$$0$2')
created build/h3_polyfill.js in 4.6s
```

I have traced the problem to this library:

```
booleanIntersects -> booleanDisjoint -> lineIntersect -> `import rbush from "geojson-rbush";` 
```

This library is used in @turf/line-intersect 6.3.0 and has the following [dependency rule](https://github.com/DenisCarriere/geojson-rbush/blob/master/package.json#L44):

```
"@turf/bbox": "*",
```

This bad rule will always install the latest version. And this is what happened because less than a month ago, a new version of turf 7.0.0 was released: https://github.com/Turfjs/turf/releases/tag/v7.0.0. This version has removed the dependency with `geojson-rbush`, but this affects existing installations using turf 6.x.

## Solution

Fix turf resolution to 6.3.0 in our package.

### DIFF
--- a/clouds/snowflake/common/package.json
+++ b/clouds/snowflake/common/package.json
@@ -1,16 +1,16 @@
 {
   "license": "BSD-3-Clause",
   "devDependencies": {
-    "@rollup/plugin-commonjs": "^17.1.0",
-    "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-node-resolve": "^13.0.0",
-    "cli-progress": "^3.11.2",
+    "@rollup/plugin-commonjs": "^26.0.1",
+    "@rollup/plugin-json": "^6.1.0",
+    "@rollup/plugin-node-resolve": "^15.2.3",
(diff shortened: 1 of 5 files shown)


vdelacruzb (approved):
LGTM!

PR #524, opened 2024-09-03
feat(sf): added warehouse option for SF

volaya:
# Description

Shortcut

- Story: https://app.shortcut.com/cartoteam/story/329793
- Autolink: [sc-329793]


vdelacruzb (approved):
LGTM. Would just add the optional parameter also in the docs in `core/clouds/snowflake/README.md` before merging.

PR #629, opened 2026-07-06
chore(rs): default Redshift Lambda runtime to python3.12 (Python 3.10 EOL)

adriantomas:
## Why

AWS Lambda ends support for **Python 3.10 on 2026-10-31** (no new functions after 2027-02-01, no updates after 2027-03-03). The Redshift CartoFunctions deploy path defaults every function to `python3.10`, so all deployed UDFs inherit the EOL runtime across the three affected accounts (shared-dev CartoFunctions / LDS lambdas, ci-resources Redshift UDFs, and data-eng). Account IDs and function inventory: internal tracker sc-559996.

## What

1. **Runtime** default `python3.10` → `python3.12` on both `cli.py` deploy paths and all `deployer.py` create/update defaults, so the codebase converges on one runtime. Per-function `clouds.redshift.config.runtime` overrides still take precedence.
2. **Build/runtime alignment**: `create_deployment_package` previously hardcoded `pip --python-version 3.10` when fetching wheels, independent of the runtime. It now takes the runtime and derives `--python-version` from it, and `deploy_function` passes its runtime through. Compiled wheels are built for the exact interpreter the function runs on, and the two values can no longer drift.

## Why python3.12 and not a newer runtime

Wheel availability was validated by replaying the deployer's exact resolver flags against every function's full pinned requirement set (transitive deps included):

`pip download -r <reqs> --platform manylinux2014_x86_64 --only-binary=:all: --python-version <v>`

- **python3.11 / python3.12** resolve all pinned deps unchanged: `numpy==1.26.4`, `scipy==1.15.3`, `shapely==2.0.6` (as cp3xx-manylinux2014 wheels), plus pure-python `geojson`/`mercantile`/`pygc`/`quadbin`/`s2sphere` and transitive `click`/`future`. numpy 1.26.4's wheels cap at cp312, so 3.12 is the newest runtime that needs no dependency changes.
- **python3.13** forces `numpy` to 2.2.6 (1.26.4 has no cp313 wheel).
- **python3.14** additionally has no `manylinux2014` numpy wheel at any version (cp314 numpy ships `manylinux_2_28` only), so it would also require raising the deployer's hardcoded `--platform` tag.

Staying on 3.12 avoids the numpy 1.x → 2.x major bump and its behavioral changes. Deployed handler code was scanned and uses no numpy-2.x-removed API, but 3.12 keeps the diff limited to the runtime string with no dependency, platform-tag, or code churn.

## Validation before merge

A validation redeploy of the CartoFunctions set on python3.12 plus a smoke test is recommended before this leaves draft.

## Notes

- Branch name still says `python311` from an earlier draft; the runtime target is python3.12.
- `PYTHON3_VERSION: 3.10.13` remains pinned across `.github/workflows/*.yml`. This is the CI runner interpreter, not the Lambda runtime; since the wheel target is now runtime-driven it no longer affects wheel correctness. Bump it separately if you want the CI tooling itself to run on a supported interpreter.

### DIFF
--- a/.claude/rules/function-dev.md
+++ b/.claude/rules/function-dev.md
@@ -61,7 +61,7 @@ clouds:
       memory_size: 512              # MB (128-10240, default 512)
       timeout: 300                  # Seconds (3-900, default 300)
       max_batch_rows: 100           # Batch size (default 100)
-      runtime: python3.10           # Python runtime
+      runtime: python3.12           # Python runtime
 ```
 
 ### Function Naming Convention
--- a/gateway/logic/clouds/redshift/cli.py
+++ b/gateway/logic/clouds/redshift/cli.py
@@ -778,7 +778,7 @@ def deploy_lambda(
 
     # Get Lambda configuration with sensible defaults
     # These can be overridden in function.yaml under clouds.redshift.config
-    runtime = cloud_config.config.get("runtime", "python3.10")
+    runtime = cloud_config.config.get("runtime", "python3.11")
     memory_size = cloud_config.config.get("memory_size", 256)  # MB
     timeout = cloud_config.config.get("timeout", 300)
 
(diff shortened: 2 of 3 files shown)


vdelacruzb (approved):
LGTM

PR #633, opened 2026-07-29
chore(bq|sf|rs|pg|db|ora): reliable dedicated/CI env cleanup and unified env naming [sc-564479]

vdelacruzb:
## Problem

Dedicated (`dedicated_*_carto`) and CI (`ci_*_carto`) environments were not reliably cleaned up:

- **Snowflake / BigQuery / Postgres / Redshift (clouds side)**: `make remove` only dropped functions/procedures one by one via `DROP_FUNCTIONS.sql`. The schema/dataset itself — plus helper procedures and any tables created by tests — was left behind. (Snowflake additionally swallows drop errors inside its JS helper procedures, so CI went green even when nothing was dropped.)
- **CI runs that failed never cleaned up**: Snowflake/BigQuery/Postgres chained `make lint && deploy && test && remove` in a single step, so any failure skipped the remove and leaked the `ci_*` schema.
- **Failures were masked**: `|| true` / best-effort echoes meant orphaned environments produced no signal.
- **Naming**: dedicated envs had three name variants (`dedicated_`, `dedicated_release_`, `dedicated_hotfix_`) with no technical purpose, and core/AT PR numbers could theoretically collide on the same `dedicated_<PR>_carto` name in the shared CD warehouses.

## Changes

- **Makefiles**: add a `drop-schema` option to the Snowflake, BigQuery, Postgres and Redshift modules `remove` targets (`DROP SCHEMA/DATASET IF EXISTS ... CASCADE`, mirroring the existing Oracle pattern; Databricks already dropped its schema). Root Makefile passes `drop-schema` explicitly to the clouds removal and no longer masks removal failures.
- **Ded workflows**: pass `drop-schema=1` on remove; remove failure masking; envs are always named `dedicated_core_<PR>_carto` — the `core_` infix keeps the namespace disjoint from the premium AT repo (`dedicated_<PR>_carto`), so one repo's PR close can never drop the other's live env. Redshift Lambda prefix stays `ded-<PR>-` (CLI enforces a 10-char limit).
- **CI workflows (SF/BQ/PG)**: split the chained step into lint / deploy / test / remove, with remove under `if: always()` and `drop-schema=1` — matching the existing Redshift/Oracle/Databricks pattern, so failed runs clean up too.
- **Docs**: `.claude/rules/` updated — CI/CD workflow table (Databricks ded workflow was missing), naming convention, always-on cleanup, and `make remove [drop-schema=1]` in every cloud's command list.

## Follow-ups (tracked in sc-564479)

- Bump the `core` submodule in CartoDB/analytics-toolbox#1121 after this merges.
- One-off sweep of pre-existing stale `dedicated_*` / `ci_*` schemas.

Shortcut: [sc-564479]

### DIFF
--- a/.claude/rules/bigquery.md
+++ b/.claude/rules/bigquery.md
@@ -29,6 +29,8 @@ cd clouds/bigquery
 make deploy   # deploy modules
 make test     # run tests (Jest)
 make build    # build JS libraries + SQL modules
+make remove   # drop deployed functions
+make remove drop-schema=1  # drop entire dataset with CASCADE (destructive)
 ```
 
 ## Key Details
--- a/.claude/rules/ci-cd.md
+++ b/.claude/rules/ci-cd.md
@@ -21,12 +21,12 @@ Each cloud has its own CI/CD workflows in `.github/workflows/`:
 | BigQuery | `bigquery.yml` | `bigquery-ded.yml` |
 | Snowflake | `snowflake.yml` | `snowflake-ded.yml` |
 | Redshift | `redshift.yml` | `redshift-ded.yml` |
-| Databricks | `databricks.yml` | - |
+| Databricks | `databricks.yml` | `databricks-ded.yml` |
 | Postgres | `postgres.yml` | `postgres-ded.yml` |
 | Oracle | `oracle.yml` | `oracle-ded.yml` |
 
 - **Main workflows**: Triggered on PRs and pushes to main. Run lint, deploy to CI env, test, cleanup.
-- **Dedicated (`-ded`) workflows**: PR-triggered, deploy to isolated environment for testing.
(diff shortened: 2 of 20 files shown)


vdelacruzb:
Added `2cbb3826`: the Snowflake/BigQuery/Postgres CI workflows now run the schema cleanup in a separate `if: always()` step (matching the existing Redshift/Oracle/Databricks pattern), so the `ci_*` schema is dropped even when lint/deploy/test fails.

vdelacruzb:
Added `2306be35`: core dedicated envs are now named `dedicated_core_<PR>_carto` (`DEDICATED_CORE_<PR>_` for Oracle). PR numbers are only unique per repo, so this keeps the namespace disjoint from the premium AT repo's `dedicated_<PR>_carto` in the shared CD warehouses — avoiding the theoretical case where one repo's PR close drops the other repo's live dedicated env. The Redshift Lambda prefix (`ded-<PR>-`) is left unchanged due to the CLI's 10-char prefix limit. Also fixed the stale CI/CD rules table (Databricks does have a ded workflow) and documented the naming convention.

miguelangelmorenochacon (approved):

PR #634, opened 2026-08-04
feat(rs|clustering): add CREATE_CLUSTERDBSCAN [sc-565418]

vdelacruzb:
Story: [sc-565418](https://app.shortcut.com/cartoteam/story/565418)

## What

Adds `CREATE_CLUSTERDBSCAN` to the Redshift `clustering` module, next to `CREATE_CLUSTERKMEANS`.

```sql
CREATE_CLUSTERDBSCAN(input, output_table, geom_column, epsilon, min_points)
CREATE_CLUSTERDBSCAN(input, output_table, geom_column, epsilon, min_points, partition_column)
```

Output is every input column plus `cluster_id` (`BIGINT`, `NULL` = noise) and `pt_type` (`core` / `border` / `noise` / `skipped`).

## Why

BigQuery has a native `ST_CLUSTERDBSCAN`. Redshift has no equivalent and the toolbox only shipped k-means for it, so density-based clustering was unavailable on Redshift entirely.

The optional `partition_column` clusters each group independently in a single call — the analogue of BigQuery's `OVER (PARTITION BY ...)`. Without it, per-group clustering means one execution per group, which does not scale.

## Why pure SQL rather than a gateway Lambda

The Lambda return leg is `VARCHAR(MAX)` = 64 KB, which caps that route at roughly 3k points — the same ceiling already documented for `CREATE_CLUSTERKMEANS` ("around 2500"). Going through Python would also require spatial tiling plus a cross-tile merge to exceed it. Pure SQL/plpgsql runs in-warehouse with no payload ceiling and less machinery.

## Semantics

Assignments match `sklearn.cluster.DBSCAN` with `metric='haversine'`. Where DBSCAN is genuinely ambiguous — a border point reachable from two clusters — the lowest canonical label wins. That is provably what scikit-learn does: it opens clusters in increasing order of minimum core index and never relabels. It also makes results reproducible across runs, which an arbitrary visit order would not.

`epsilon` is in meters and maps to BigQuery's `epsilon`; `min_points` maps to BigQuery's `minimum_geographies`. `min_points` is used rather than `minimum_geographies` because this procedure accepts points only.

## Implementation notes for review

Three parts are non-obvious and worth a look:

1. **Neighbour pre-filter.** Candidates are narrowed with a lon/lat grid whose cell size comes from the haversine identity, so the 3x3 ring is provably sufficient for any `epsilon` and latitude. Deriving it from a linear "meters per degree" approximation instead would understate the longitude bound by about `(epsilon/2R)^2/6 * tan(lat)^2` and silently drop pairs. The exact `ST_DistanceSphere` test still runs on every surviving pair, so the filter can only add candidates, never remove true neighbours. `gx_span >= 3` is load-bearing: below that the ring repeats cells and emits duplicate edges, which would corrupt the degree count.
2. **Pointer jumping in the propagation loop.** Min-label propagation plus `comp := comp[comp]` reaches the same fixpoint in O(log n) passes rather than one pass per link. This is not just a speedup — without it a chain longer than the iteration guard never converges.
3. **All statements are dynamic SQL.** The internal temp tables are dropped and recreated on every call, so static SQL risks a stale cached plan on the second call in a session. Edges are also built one direction and mirrored, making the symmetry the components fixpoint relies on structural rather than assumed.

## Validation

Run against a Redshift dev cluster with scikit-learn as the oracle. Comparisons are label-invariant (groupings, not raw ids).

| Check | Result |
|---|---|
| 12 hand-built edge cases | identical to scikit-learn |
| 30 randomized fuzz trials (blobs, uniform, chains, duplicates, polar 80-88, antimeridian) | identical to scikit-learn |
| 3,000-point mixed set (2,624 core / 56 border / 320 noise) | identical, including every border point |
| `ST_DistanceSphere` radius | measured 6371008.8 m, matches the constant used |
| Grid pre-filter disabled | byte-identical output |
| Pointer jumping disabled (guard raised) | byte-identical output |
| Propagation passes, chains of 60/120/300/600 | 6/7/9/10 with jumping vs 58/118/298/598 without |

Performance:

| Workload | Time |
|---|---|
| 1,000,000 points / 100,000 partitions | ~7 s |
| 500,000 points / 50,000 partitions | ~7 s |
| 20,000 points, one dense partition (~1,000 neighbours each) | ~5 s |

Lint clean: `sqlfluff`, `brunette`, `flake8`, `markdownlint`.

## Test notes

`modules/test/clustering/test_CREATE_CLUSTERDBSCAN.py` keeps 6 cases, in the style of the existing module tests: happy path (also called twice to cover the same-session regression), core/border/noise including the case that separates DBSCAN from single-linkage, NULL geometry, `partition_column`, antimeridian + high latitude, and a 300-point chain for convergence. The broader differential and fuzz testing was run locally against scikit-learn and is deliberately not committed, since it needs scikit-learn and long runtimes.

The 4 pre-existing `CLUSTERKMEANS` tests fail in my environment with `Error assuming specified role` from the Lambda external function. Unrelated to this change — this procedure calls no external functions.

## Known limitations (documented in the doc page)

- `POINT` geometries only. `ST_DistanceSphere` accepts nothing else, so other types are rejected rather than silently approximated by centroids.
- Rows with a `NULL` geometry or `NULL` partition value are excluded and reported as `pt_type = 'skipped'`. This differs from SQL `PARTITION BY` semantics, where `NULL` forms its own group, and therefore from BigQuery for NULL partition keys. Flagging it explicitly in case reviewers prefer grouping them.
- Cost scales with point density rather than row count.

### DIFF
--- a/clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
+++ b/clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
@@ -91,7 +91,7 @@ BEGIN
     -- cluster_id, pt_type or __carto_idx would produce a duplicate column name.
     -- The obvious way to hit this is re-running the procedure on its own output,
     -- so fail with a clear message instead of a raw duplicate-column error.
-    -- Materialising zero rows lets this work for a subquery input too.
+    -- Materializing zero rows lets this work for a subquery input too.
     EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_cols';
     EXECUTE 'CREATE TEMP TABLE __carto_dbscan_cols AS
              SELECT * FROM ' || input_query || ' LIMIT 0';
@@ -118,9 +118,8 @@ BEGIN
         -- The partition key is compared with an equality join, which never
         -- matches NULL to NULL. Cast it to VARCHAR and substitute a sentinel so
         -- that rows with a NULL partition value form their own group, which is
-        -- what SQL PARTITION BY / GROUP BY do and what BigQuery's
-        -- ST_CLUSTERDBSCAN(...) OVER (PARTITION BY ...) would do. Excluding them
-        -- instead would silently drop rows from the analysis.
+        -- what SQL PARTITION BY and GROUP BY both do. Excluding them instead
+        -- would silently drop rows from the analysis.
         part_expr := 'COALESCE(' || partition_column ||
                      '::VARCHAR, ''__carto_null_partition__'')';
         dist_col  := 'part';
@@ -147,7 +146,7 @@ BEGIN
         FROM ' || input_query;
 
     ------------------------------------------------------------------
-    -- 2. Grid parameters for the neighbour pre-filter
+    -- 2. Grid parameters for the neighbor pre-filter
     ------------------------------------------------------------------
     -- dlat/dlon are sized so that any two points within epsilon fall in the same
     -- or an adjacent cell. dlon is computed at the HIGHEST |latitude| present,
@@ -206,7 +205,7 @@ BEGIN
         WHERE ' || geom_column || ' IS NOT NULL';
 
     ------------------------------------------------------------------
-    -- 4. Probe cells: each point claims its own cell and its 8 neighbours
+    -- 4. Probe cells: each point claims its own cell and its 8 neighbors
     ------------------------------------------------------------------
     -- Expanding the 3x3 ring on one side turns the range predicate into an
     -- equijoin, so step 5 is a hash join rather than a nested loop.
@@ -318,12 +317,12 @@ BEGIN
     END IF;
 
     ------------------------------------------------------------------
-    -- 8. Border points: smallest component label among core neighbours
+    -- 8. Border points: smallest component label among core neighbors
     ------------------------------------------------------------------
     -- MIN over canonical component labels reproduces sklearn exactly: sklearn
     -- opens clusters in increasing order of minimum core index and never
     -- relabels, so a border point joins the cluster with the smallest
-    -- minimum-core-index among its core neighbours.
+    -- minimum-core-index among its core neighbors.
     EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_border';
     EXECUTE 'CREATE TEMP TABLE __carto_dbscan_border DISTKEY(idx) SORTKEY(idx) AS
         SELECT e.a AS idx, MIN(cc.comp) AS comp
@@ -336,9 +335,9 @@ BEGIN
     ------------------------------------------------------------------
     -- 9. Write labels back
     ------------------------------------------------------------------
-    -- cluster_id is 0-based and dense per partition, matching sklearn labels_
-    -- and BigQuery ST_CLUSTERDBSCAN. Noise is NULL (BigQuery convention;
-    -- sklearn uses -1).
+    -- cluster_id is 0-based and dense per partition, matching the reference
+    -- DBSCAN labelling. Noise is NULL rather than a sentinel such as -1, so it
+    -- reads as SQL's absence of a value.
     EXECUTE 'UPDATE ' || output_table || ' SET
                 cluster_id = s.cluster_id,
                 pt_type    = s.pt_type
--- a/clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
+++ b/clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
@@ -57,23 +57,79 @@ BEGIN
     -- Trim first: a padded table name would otherwise match the whitespace test
     -- and be wrapped in parentheses as though it were a query.
     input_query := BTRIM(input);
-    EXECUTE 'SELECT regexp_count(''' || input_query || ''', ''\\\\s'')' INTO table_format;
+    -- Evaluated directly rather than through EXECUTE: interpolating the input
+    -- into a SQL literal breaks on any query containing a quote, which rules
+    -- out most real queries (WHERE country = 'ES').
+    table_format := REGEXP_COUNT(input_query, '\\s');
     IF table_format > 0
     THEN
         input_query := '(' || input_query || ')';
     END IF;
 
     -- Validate output table
-    EXECUTE 'SELECT split_part(''' || output_table || ''', ''.'', 1)' INTO output_first;
-    EXECUTE 'SELECT split_part(''' || output_table || ''', ''.'', 2)' INTO output_second;
-    EXECUTE 'SELECT split_part(''' || output_table || ''', ''.'', 3)' INTO output_third;
-    EXECUTE 'SELECT split_part(''' || output_table || ''', ''.'', 4)' INTO output_fourth;
+    output_first  := SPLIT_PART(output_table, '.', 1);
+    output_second := SPLIT_PART(output_table, '.', 2);
+    output_third  := SPLIT_PART(output_table, '.', 3);
+    output_fourth := SPLIT_PART(output_table, '.', 4);
     IF output_first = '' OR output_second = '' OR output_fourth != ''
     THEN
         output_table := 'Invalid output table name. It must have the form [DATABASE.]SCHEMA.TABLE';
         RAISE INFO 'Invalid output table name. It must have the form [DATABASE.]SCHEMA.TABLE';
         RETURN;
     END IF;
+
+    -- Column checks, all against a zero-row materialization of the input so they
+    -- work for a subquery as well as a table name. Doing them before touching the
+    -- data means a wrong column name is reported plainly instead of surfacing as
+    -- a raw error that quotes internal SQL.
+    EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_cols';
+    EXECUTE 'CREATE TEMP TABLE __carto_dbscan_cols AS
+             SELECT * FROM ' || input_query || ' LIMIT 0';
+
+    -- The output is built with SELECT *, so an input that already carries
+    -- cluster_id, pt_type or __carto_idx would produce a duplicate column name.
+    -- The obvious way to hit this is re-running the procedure on its own output.
+    EXECUTE 'SELECT COUNT(*) FROM pg_attribute a
+             JOIN pg_class c ON c.oid = a.attrelid
+             WHERE c.relname = ''__carto_dbscan_cols'' AND a.attnum > 0
+               AND LOWER(a.attname) IN
+                   (''cluster_id'', ''pt_type'', ''__carto_idx'')' INTO bad_cols;
+    IF bad_cols > 0
+    THEN
+        EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_cols';
+        output_table := 'Invalid input. It must not contain columns named cluster_id, pt_type or __carto_idx';
+        RAISE INFO 'Invalid input. It must not contain columns named cluster_id, pt_type or __carto_idx';
+        RETURN;
+    END IF;
+
+    EXECUTE 'SELECT COUNT(*) FROM pg_attribute a
+             JOIN pg_class c ON c.oid = a.attrelid
+             WHERE c.relname = ''__carto_dbscan_cols'' AND a.attnum > 0
+               AND LOWER(a.attname) = LOWER(''' || BTRIM(geom_column) || ''')' INTO bad_cols;
+    IF bad_cols = 0
+    THEN
+        EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_cols';
+        output_table := 'Invalid geom_column. It does not exist in input';
+        RAISE INFO 'Invalid geom_column. It does not exist in input';
+        RETURN;
+    END IF;
+
+    IF partition_column IS NOT NULL AND BTRIM(partition_column) != ''
+    THEN
+        EXECUTE 'SELECT COUNT(*) FROM pg_attribute a
+                 JOIN pg_class c ON c.oid = a.attrelid
+                 WHERE c.relname = ''__carto_dbscan_cols'' AND a.attnum > 0
+                   AND LOWER(a.attname) = LOWER(''' || BTRIM(partition_column) || ''')' INTO bad_cols;
+        IF bad_cols = 0
+        THEN
+            EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_cols';
+            output_table := 'Invalid partition_column. It does not exist in input';
+            RAISE INFO 'Invalid partition_column. It does not exist in input';
+            RETURN;
+        END IF;
+    END IF;
+
+    EXECUTE 'DROP TABLE IF EXISTS __carto_dbscan_cols';
 
     -- ST_DistanceSphere reads coordinates as degrees and only accepts points.
     -- SRID 0 is tolerated (lon/lat without a declared SRID is common); anything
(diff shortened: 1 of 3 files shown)


vdelacruzb:
## Acceptance: cross-checked against BigQuery's native `ST_CLUSTERDBSCAN`

Validated this procedure against BigQuery's native implementation on the same 1.4M row POI dataset, since BigQuery is the reference for DBSCAN in the toolbox.

### Setup

```sql
-- Redshift (this PR)
CALL carto.CREATE_CLUSTERDBSCAN(
    'demo_tables.osm_pois_usa',
    'vdelacruz_carto.osm_pois_dbscan',
    'geom', 100, 5, 'group_name');
```

```sql
-- BigQuery (native, reference)
CREATE OR REPLACE TABLE `cartodb-data-engineering-team.vdelacruz_carto.osm_pois_dbscan` AS
SELECT
  osm_id, name, address, group_name, subgroup_name, geom,
  ST_CLUSTERDBSCAN(geom, 100, 5) OVER (PARTITION BY group_name) AS cluster_id
FROM `carto-demo-data.demo_siteselection.osm_pois_usa`;
```

`epsilon` maps to BigQuery's `epsilon`, `min_points` to `minimum_geographies` (verified empirically that BigQuery counts the geography itself), and `partition_column` to `OVER (PARTITION BY ...)`.

The two source tables live in different places, so they were confirmed to hold the same data first: `COUNT(*)`, `SUM(osm_id)`, `SUM(ROUND(lon,7))` and `SUM(ROUND(lat,7))` per `group_name` match exactly across all 11 groups.

### Totals

```sql
-- Redshift
SELECT COUNT(*) AS total_rows,
       COUNT(DISTINCT group_name || ':' || cluster_id) AS clusters,
       SUM(CASE WHEN cluster_id IS NOT NULL THEN 1 ELSE 0 END) AS clustered,
       SUM(CASE WHEN cluster_id IS NULL THEN 1 ELSE 0 END)     AS unclustered
FROM vdelacruz_carto.osm_pois_dbscan;

-- BigQuery
SELECT COUNT(*) AS total_rows,
       COUNT(DISTINCT CONCAT(group_name,':',CAST(cluster_id AS STRING))) AS clusters,
       COUNTIF(cluster_id IS NOT NULL) AS clustered,
       COUNTIF(cluster_id IS NULL)     AS unclustered
FROM `cartodb-data-engineering-team.vdelacruz_carto.osm_pois_dbscan`;
```

| | total_rows | clusters | clustered | unclustered |
|---|---|---|---|---|
| Redshift | 1,437,379 | **20,683** | 243,606 | 1,193,773 |
| BigQuery | 1,437,379 | **20,683** | 243,605 | 1,193,774 |

Identical cluster count; one row differs.

### Cluster membership, compared exactly

Cluster ids are not comparable directly — both sides emit dense zero-based labels but number them in different visit orders. Membership was compared instead, via an MD5 over each cluster's sorted member ids. `MD5` is byte-identical across the two engines (both return `9d7f94e747ed612a2b08e63092e51a14` for `'101,102'`), so the fingerprint sets can be compared directly.

```sql
-- Redshift
SELECT group_name, MD5(LISTAGG(osm_id::VARCHAR,',') WITHIN GROUP (ORDER BY osm_id)) AS sig
FROM vdelacruz_carto.osm_pois_dbscan WHERE cluster_id IS NOT NULL
GROUP BY group_name, cluster_id;

-- BigQuery
SELECT group_name, TO_HEX(MD5(STRING_AGG(CAST(osm_id AS STRING),',' ORDER BY osm_id))) AS sig
FROM `cartodb-data-engineering-team.vdelacruz_carto.osm_pois_dbscan` WHERE cluster_id IS NOT NULL
GROUP BY group_name, cluster_id;
```

| Measure | Result |
|---|---|
| Clusters with byte-identical membership | **20,292 of 20,683 (98.1%)** |
| Clusters differing | 391 on each side |
| Rows assigned to a different cluster | **843 of 242,751 — 0.35% of clustered rows, 0.059% of input** |
| Rows clustered by one engine only | **1** |

### Why the remaining 0.059% differs

The two engines measure on spheres of slightly different radius: BigQuery uses **6,371,010 m**, Redshift's `ST_DistanceSphere` **6,371,008.8 m**. That is 0.000019%, about **19 µm at a 100 m radius**, so it only flips pairs sitting within ~19 µm of the threshold. A single flipped edge can merge or split a whole connected component and relabel every member, which is why 471 of the differing rows are core points rather than only border points.

The one row clustered by Redshift alone is a clean illustration. POI `3253163797` ("Luna Grill") has a same-group neighbour at:

- **99.999993 m** by `ST_DistanceSphere` (Redshift) — inside `epsilon`, so it joins the cluster as a border point
- **100.0000127 m** by `ST_DISTANCE` (BigQuery) — outside `epsilon`, so it is noise

Predicting the BigQuery value from the radius ratio gives `99.999993 × (6371010 / 6371008.8) = 100.0000118`, matching the measured value to seven decimals.

### Independent confirmation

To separate "different distance function" from "different algorithm", the full `Healthcare` group (34,415 points, no windowing) was also clustered with `sklearn.cluster.DBSCAN` using `metric='haversine'` at Redshift's radius:

- 232 clusters on both sides, **all 232 with identical membership**
- identical noise sets (32,517 points)

So the implementation agrees exactly with the reference algorithm under its own metric, and the BigQuery deltas are attributable to the distance function alone.

### Notes for anyone repeating this comparison

- Comparing per-group `COUNT` and `SUM(id)` is **not** sufficient. Border points moving between clusters inside the same group leave both unchanged; that screen passed 9 of 10 groups while 391 clusters actually differed. Use the per-cluster fingerprint.
- `osm_id` is not unique in this dataset (3,569 duplicate rows over 2,928 repeated ids), so any row-level join keyed on it silently collapses rows.
- Redshift `LISTAGG` caps at 65,535 characters. It fits here (largest cluster 3,421 members), but a larger `epsilon` may overflow it.

clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
```diff
@@ -0,0 +1,469 @@
+--------------------------------
+-- Copyright (C) 2026 CARTO
+--------------------------------
+
+CREATE OR REPLACE PROCEDURE @@RS_SCHEMA@@.CREATE_CLUSTERDBSCAN
+(
+    input VARCHAR(MAX),
+    output_table INOUT VARCHAR(MAX),
```
dasamu, line 8: `output_table` being `INOUT` makes this procedure uncallable from the CARTO Workflows **Call Procedure** component.

Redshift requires an `INOUT` argument to be bound to a **writable variable**. The Workflows engine wraps the user's SQL inside a generated plpgsql procedure and substitutes the output table as a **string literal**, so the call fails:

```
ERROR 42601: argument 2 is an output argument but is not writable
  (at: PL/pgSQL function "_temp_c992da0a8181de15" line 8 at call)
```

Reproduced in isolation on a live cluster — the only difference between these two is literal vs variable:

```sql
-- fails
CREATE OR REPLACE PROCEDURE __repro() AS $$
BEGIN
  CALL carto.CREATE_CLUSTERDBSCAN('public.pts','public.out','geom',25,3,'id');
END; $$ LANGUAGE plpgsql;
-- ERROR: argument 2 is an output argument but is not writable

-- works
CREATE OR REPLACE PROCEDURE __fix() AS $$
DECLARE msg VARCHAR(MAX);
BEGIN
  msg := 'public.out';
  CALL carto.CREATE_CLUSTERDBSCAN('public.pts', msg, 'geom', 25, 3, 'id');
END; $$ LANGUAGE plpgsql;
-- OK
```

This matters because a Workflow is the primary way most customers would reach this procedure — the partitioned batch case the description motivates ("one execution per group, which does not scale") is exactly a Workflow use case.

I worked around it with a thin `IN`-only wrapper and verified it end to end in a real Workflow: correct output, and the parameters demonstrably reach the procedure (`epsilon` 25→100 with `min_points` 3→5 moved the result from 244 clusters / 2,827 noise to 120 / 3,001, matching a direct call exactly).

```sql
CREATE OR REPLACE PROCEDURE carto.CREATE_CLUSTERDBSCAN_WF(
    input VARCHAR(MAX), output_table VARCHAR(MAX), geom_column VARCHAR(MAX),
    epsilon FLOAT8, min_points INT, partition_column VARCHAR(MAX))
AS $$
DECLARE msg VARCHAR(MAX);
BEGIN
    msg := output_table;
    CALL carto.CREATE_CLUSTERDBSCAN(input, msg, geom_column, epsilon, min_points, partition_column);
    IF msg NOT LIKE 'Table %'
    THEN
        RAISE EXCEPTION 'CREATE_CLUSTERDBSCAN failed: %', msg;
    END IF;
END;
$$ LANGUAGE plpgsql;
```

The `RAISE EXCEPTION` matters independently of the `INOUT` question: on a validation failure the procedure sets `output_table` to a message and returns without creating the table, so a Workflow node would otherwise report **success** while producing nothing.

Options, roughly in order of preference:

1. Ship an `IN`-only variant alongside the `INOUT` one, so Workflows can call it directly.
2. Keep `INOUT` and document the wrapper pattern on the doc page.
3. Leave as-is and accept that the procedure is SQL-client-only.

Not specific to this PR — `CREATE_CLUSTERKMEANS` inherits the same convention and presumably has the same limitation, so it may be worth deciding once for the module.

clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
```diff
@@ -0,0 +1,469 @@
...
+        output_table := 'Invalid min_points. It must be >= 1';
+        RAISE INFO 'Invalid min_points. It must be >= 1';
+        RETURN;
+    END IF;
+
+    -- Accept either a table name or a subquery, as CREATE_CLUSTERKMEANS does.
+    -- Trim first: a padded table name would otherwise match the whitespace test
+    -- and be wrapped in parentheses as though it were a query.
+    input_query := BTRIM(input);
+    -- Evaluated directly rather than through EXECUTE: interpolating the input
```
dasamu, line 60: This breaks for any `input` containing a single quote, because the value is concatenated into a string literal unescaped:

```sql
CALL carto.CREATE_CLUSTERDBSCAN(
  'SELECT * FROM public.visits WHERE id_cte LIKE ''1-1-615%''',
  'public.out', 'geom', 25, 3, 'id_cte');
```

```
ERROR: syntax error at or near "1" in context
  "SELECT regexp_count('SELECT * FROM public.visits WHERE id_cte LIKE '1"
```

A quote-free subquery works fine (verified — same 244 clusters as a table-name input), so this is specifically **subqueries containing a string literal**. Since the comment directly above advertises subquery input, and `WHERE ... LIKE '...'` or `WHERE country = 'ES'` is the most natural subquery a user would pass, I'd treat this as a functional bug rather than only the injection surface it resembles.

The same pattern is used just below for `output_table` (the four `split_part` calls).

Neither needs `EXECUTE` at all — binding the variables directly avoids the round-trip and the escaping entirely. Verified on Redshift, including with a quoted input:

```sql
SELECT regexp_count(input_query, '\\s') INTO table_format;
SELECT split_part(output_table, '.', 1), split_part(output_table, '.', 2),
       split_part(output_table, '.', 3), split_part(output_table, '.', 4)
  INTO output_first, output_second, output_third, output_fourth;
```

Also inherited from `CREATE_CLUSTERKMEANS`, so likely pre-existing there too.

dasamu (commented):
Two integration issues found while testing this on a live Redshift Serverless cluster — details inline.

clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
vdelacruzb: Good catch, and the diagnosis is exactly right — but this landed in `cdabffc7`, which is the commit this comment is pinned to, so I think you hit a copy deployed to your schema before that push. Both `EXECUTE` sites are gone:

```sql
table_format  := REGEXP_COUNT(input_query, '\\s');
output_first  := SPLIT_PART(output_table, '.', 1);
output_second := SPLIT_PART(output_table, '.', 2);
output_third  := SPLIT_PART(output_table, '.', 3);
output_fourth := SPLIT_PART(output_table, '.', 4);
```

Same idea as your `SELECT ... INTO`, just as plain assignments. `grep -nE 'regexp_count|split_part'` on the file now returns nothing.

Re-deployed current `HEAD` and ran your repro shape against a table with your `id_cte` values:

```sql
CALL carto.CREATE_CLUSTERDBSCAN(
  'SELECT * FROM vdelacruz_carto.visits WHERE id_cte LIKE ''1-1-615%''',
  'vdelacruz_carto.visits_out', 'geom', 25, 3, 'id_cte');
-- Table vdelacruz_carto.visits_out created with the clustering
-- 8 rows, 1 cluster
```

Also passing: `WHERE id_cte = '1-1-615'`, and two literals in one predicate (`id_cte <> 'zzz' AND id_cte LIKE '%-615'`). There is a regression test for it in `test_create_clusterdbscan_partition_column`, which passes a subquery deriving its partition column with `CASE ... THEN ''low''`.

Your framing that it is a functional bug rather than just an injection surface is what prompted the change — the comment above the code advertises subquery input, so failing on `WHERE country = 'ES'` made the feature mostly unusable. Worth noting the fix also removes five dynamic statements.

On `CREATE_CLUSTERKMEANS` inheriting it: agreed, it still has the original pattern. Out of scope here, but happy to open a follow-up.

dasamu: You're right, and thanks for the detail. I tested a copy fetched at ~09:00Z; `cdabffc7` landed at 09:16Z and I posted at 09:48Z, so I reviewed code that was already 32 minutes stale. My mistake — I should have re-fetched before posting.

Verified your fix on `cdabffc7` against a live cluster. All three shapes create the table:

```sql
WHERE id_cte LIKE '1-1-615%'                    -- 80 rows, 3 clusters
WHERE id_cte <> 'zzz' AND id_cte LIKE '%-615%'  -- two literals in one predicate
WHERE id_cte = '1-1-615-57965'                  -- equality
```

Regression is clean too — the real-data sample still gives 244 clusters with core 1,061 / border 13 / noise 2,827, zero mismatches against my reference implementation (compared label-invariantly, so numbering differences can't mask a real disagreement). The column checks added in the same commit all behave as intended: re-running the procedure on its own output, a bad `geom_column`, and a bad `partition_column` each return their plain message rather than a raw SQL error.

Agreed on `CREATE_CLUSTERKMEANS` being out of scope here — a follow-up sounds right.

One note: the `INOUT` finding in the other thread does still reproduce on `cdabffc7`, so that one isn't affected by this fix.

clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
vdelacruzb: Thanks — the repro is accurate, but I'd keep `INOUT` here, because it's the module convention and the engine already has a pattern for it. Also: it **is** callable from Workflows today. Details below.

### `INOUT` is the convention

20 Redshift procedures in the toolbox declare `output_table INOUT VARCHAR(MAX)` — `CREATE_CLUSTERKMEANS`, `CREATE_ROUTES`, `CREATE_ISOLINES`, the tiler ones. Changing it here would make this the odd one out rather than fix anything.

### The engine already binds a variable

`workflows-engine/src/components/createroutes.ts`, Redshift branch:

```ts
DECLARE
  outputTableName VARCHAR(MAX):= '${tablename}';
BEGIN
CALL carto.CREATE_ROUTES(
  'SELECT ... FROM ${source}',
  outputTableName,          // variable, not a literal
  ...
```

`createroutes.ts` has 3 `DECLARE`s; the generic `call.ts` has none. So what you hit is a limitation of the generic **Call Procedure** component, and it applies to all 20 `INOUT` procedures equally — not something this PR introduces. A dedicated component for this procedure would follow `createroutes.ts` and work unchanged.

### How to call it from Workflows today

The generic component does work, because it places your SQL inside a nested `BEGIN … END;` in a generated plpgsql procedure, so a nested `DECLARE` block is valid. In the **Call Procedure** component's SQL box:

```sql
DECLARE
  out_table VARCHAR(MAX) := '$output';
BEGIN
  CALL carto.CREATE_CLUSTERDBSCAN('$a', out_table, 'geom', 100, 5, 'group_name');
  IF out_table NOT LIKE 'Table %' THEN
    RAISE EXCEPTION 'CREATE_CLUSTERDBSCAN failed: %', out_table;
  END IF;
END;
```

`$a` and `$output` are substituted as bare table names — the component deliberately leaves quoting to the caller — so the quotes around them are yours to add.

Verified by emulating the connection's `wrapCodeInIfTableExistsBlock` output on a live cluster:

| Pattern | Result |
|---|---|
| `CALL ... ('$a', '$output', ...)` | `argument 2 is an output argument but is not writable` — your error |
| nested `DECLARE` block above | works, output table written |
| same with `epsilon = 0` | raises `CREATE_CLUSTERDBSCAN failed: Invalid epsilon. It must be a positive distance in meters` |

### Your second point is the portable one

The status check above is worth keeping regardless of the calling convention. On a validation failure the procedure sets `output_table` to a message and returns **without creating the table**, so any caller that ignores that message sees success with no output. That's inherent to the `INOUT`-message convention shared by all 20 procedures rather than specific to this one, but it is a real silent-failure mode and your `RAISE` is the right guard. Worth a separate ticket to either teach the generic component the `DECLARE` pattern or have the dedicated components assert on the returned message — happy to open it.

No change in this PR, then, unless you disagree.

clouds/redshift/modules/sql/clustering/CREATE_CLUSTERDBSCAN.sql
vdelacruzb: Just for the record. A discussed in slack and documented properly for workflows. The `INOUT` issues can be solved by passing a variable as output_table. In this PR is introduced for consistency with the rest of procedures in the AT.

diego-vicente (approved):
The algorithm looks solid, LGMT

PR #636, opened 2026-08-04
fix(sf): unblock Snowflake CI — pin @azure/storage-common and tolerate H3_POLYFILL release drift [sc-565492][sc-565533]

vdelacruzb:
## Problem

Two independent issues currently break Snowflake CI, and they block each other's PRs (any Snowflake-touching PR needs both to go green), so they ship together here — one commit per issue, tracked by separate tickets.

### 1. `@azure/storage-common@12.5.0` breaks every fresh yarn install (sc-565492)

The transitive dep (via `snowflake-sdk` → `@azure/storage-blob`) declares `engines.node >= 22`; CI runs Node 20.19. `yarn.lock` is not tracked, so every run re-resolves and fails. Same class of breakage as #630. Also breaks the premium AT repo (picks up the fix via its next core submodule bump).

**Fix**: pin `"@azure/storage-common": "12.4.1"` in the existing `resolutions` block. Verified with a clean install using the exact CI command (`yarn -s --update-checksums`).

### 2. `H3_POLYFILL(africa, 3, 'contains')` returns 2177 vs expected 2178 (sc-565533)

Deterministic on the CI account, while a dev account on Snowflake `10.26.102` returns 2178 with byte-identical code (verified by fresh-deploying `main` there). The candidate set is stable (center mode still passes at 2323); the drift is in what `ST_ASGEOJSON(ST_BUFFER(...))` feeds the JS `booleanContains` filter — a float-borderline edge cell flips between Snowflake releases.

**Fix**: the africa/contains fixture accepts `[2177, 2178]` (all other fixtures stay exact), keeping the suite green on both sides of the release rollout.

Replaces #638 (closed; its commit is cherry-picked here unchanged).

Shortcut: [sc-565492] [sc-565533]

### DIFF
--- a/clouds/snowflake/common/package.json
+++ b/clouds/snowflake/common/package.json
@@ -2,6 +2,7 @@
   "license": "BSD-3-Clause",
   "resolutions": {
     "@azure/storage-blob": "12.32.0",
+    "@azure/storage-common": "12.4.1",
     "@azure/core-paging": "1.6.2",
     "@azure/logger": "1.3.0",
     "@azure/core-xml": "1.5.1",
--- a/clouds/snowflake/modules/test/h3/H3_POLYFILL.test.js
+++ b/clouds/snowflake/modules/test/h3/H3_POLYFILL.test.js
@@ -39,10 +39,11 @@ describe('H3_POLYFILL should support POLYGON geographies', () => {
         ['spain', 6, 'center', 13344],
         ['spain', 6, 'intersects', 13701],
         ['africa', 3, undefined, 2323],
-        ['africa', 3, 'contains', 2178],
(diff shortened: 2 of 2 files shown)


vdelacruzb:
The next action was failing: https://github.com/CartoDB/analytics-toolbox-core/actions/runs/30898219379
The fix works when adding the dedicated label and removing it.

vdelacruzb:
The H3_POLYFILL failure blocking this PR is unrelated Snowflake release drift — root cause and fix in #638 (sc-565533). Once #638 merges, re-running CI here should go green: the azure pin itself is already proven by this PR's run (install/lint/deploy/remove all passed; only the H3 fixture failed).

miguelangelmorenochacon (approved):

PR #638, opened 2026-08-04
fix(sf|h3): tolerate Snowflake release drift in H3_POLYFILL contains-mode africa fixture [sc-565533]

vdelacruzb:
## Problem

`H3_POLYFILL(africa, 3, 'contains')` fails deterministically in CI: expected 2178 cells, received 2177 (confirmed twice on https://github.com/CartoDB/analytics-toolbox-core/actions/runs/30916924966). This currently blocks every Snowflake-touching PR, including #636.

## Root cause — Snowflake release drift, not a code change

The SQL (`H3_POLYFILL.sql`, unchanged since Apr 2024) builds the candidate set with native `H3_POLYGON_TO_CELLS_STRINGS` (center mode — still passes at 2323, so the candidate set is stable) and filters with the bundled turf `booleanContains` (JS lib unchanged since Oct 2025) on `ST_ASGEOJSON(ST_BUFFER(...))` output.

Verification: fresh-deploying current `main` to a dev account on Snowflake `10.26.102` returns **2178**, while the CI account returns **2177** — byte-identical code, two Snowflake releases, one float-borderline edge cell flipping between contained/not-contained (most likely a serialization/precision change feeding the JS filter).

## Fix

Accept `[2177, 2178]` for this one fixture (the assertion stays exact for every other case), so the suite is green on both sides of the release rollout. If Snowflake later stabilizes on one value, the fixture can be re-pinned.

Shortcut: [sc-565533]

### DIFF
--- a/clouds/snowflake/modules/test/h3/H3_POLYFILL.test.js
+++ b/clouds/snowflake/modules/test/h3/H3_POLYFILL.test.js
@@ -39,10 +39,11 @@ describe('H3_POLYFILL should support POLYGON geographies', () => {
         ['spain', 6, 'center', 13344],
         ['spain', 6, 'intersects', 13701],
         ['africa', 3, undefined, 2323],
-        ['africa', 3, 'contains', 2178],
+        // A float-borderline edge cell makes the contains count vary across Snowflake releases
+        ['africa', 3, 'contains', [2177, 2178]],
         ['africa', 3, 'center', 2323],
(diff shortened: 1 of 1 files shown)


vdelacruzb:
Combined into #636 — the two fixes block each other's CI (this PR's yarn install fails without the azure pin, and #636's tests fail without this fixture change). The commit here is cherry-picked into #636 unchanged; tickets stay split (sc-565533 / sc-565492).

PR #649, opened 2026-09-03
chore(sf): ship JavaScript source maps in the native app package [sc-571986]

vdelacruzb:
## Why

Patch 27 of the Snowflake Native App package (AT v1.26.2) was **rejected** by Snowflake's automated security scan:

> Application code includes minified Javascript code. Snowflake security policy requires all application code to be un-obfuscated, meaning that the code must be human readable. This requirement includes minified JavaScript code.

Nothing changed on our side — `terser()` has been in this rollup chain since 2022 (`8e7f813d`), and both the accepted patch 26 and the rejected patch 27 inline the same set of libraries. Snowflake's [security requirements](https://docs.snowflake.com/en/developer-guide/native-apps/security-app-requirements) document the compliant path:

> If an app needs to use minified JavaScript code, it must include a corresponding source map file that can be used to recover the un-minified code.

and Snowflake support then asked for:

> the `.js.map` naming convention, and as far as possible name the file(s) in a way that makes it easy to map to the files (and if applicable, functions) where you use the minified js

The alternative is the appeal process — one appeal per patch, 3-5 business days, no carryover — which does not work against a monthly release cadence. **Installed functions stay minified**; we ship maps so the un-minified source is recoverable.

## What changes

**`common/rollup.config.js`** — emits a source map per library. `sourcesContent` is kept, so the un-minified code is recoverable from the map alone. `sourcemapPathTransform` re-roots the source paths at the repository directory (`analytics-toolbox/clouds/snowflake/…`, `core/clouds/snowflake/…`) instead of leaking the build machine's layout.

**`common/build_source_review.js`** (new) — generates `SOURCE_REVIEW.md` and validates the result. Snowflake JS UDFs have no import mechanism, so each library is inlined into every function that uses it: 39 libraries across 96 functions, `data_enrichment` alone into 17. One map therefore covers several functions, so the index answers the question a reviewer actually has — *given this function, where is its source* — as a `function → map` table:

```
| Function | Source map |
| --- | --- |
| `_GET_HISTOGRAM_NBINS` | `sourcemaps/statistics.js.map` |
| `_LINEAR_REGRESSION`   | `sourcemaps/statistics.js.map` |
```

An earlier revision also carried per-library sizes, bundle SHA-256s and build provenance. Those were removed deliberately after review of the customer-facing version: in-package hashes are self-referential, and the reproducibility they supported does not work for a reviewer who cannot rebuild a private repo. The document is now the review path and nothing else.

**The same script is the guard.** It fails the build when an inlined library has no map, when a map cannot yield every one of its originals (no sources at all, a length mismatch, or any null/empty entry), when a map is not valid JSON, when a bundle has no `sourceMappingURL`, or when a map leaks build paths. `--source_maps_dir` is required rather than defaulted, so `APP_SOURCE_MAPS_DIR` is the only place the directory is named. Without it a future release could silently ship unmapped JavaScript and be rejected again — and it earned its keep immediately: it caught my first `sourcemapPathTransform`, which worked out-of-tree but was a no-op under the real build.

**Makefiles** — `sourcemaps/` in the package, the `sourceMappingURL` prefix applied before the bundles are inlined, and a second `PUT` for the sub-directory.

## Package shape

```
manifest.yml, README.md, setup_script.sql, get_modules_sql_from_stage.py   unchanged
modules.sql          minified as before + //# sourceMappingURL=sourcemaps/<lib>.js.map
SOURCE_REVIEW.md     the index (kept at the root so a reviewer finds it)
sourcemaps/       39 × <lib>.js.map
```

The shipped SQL is otherwise unchanged: the only difference inside each function body is the trailing comment.

## Verification

Built the premium package end to end (`make build-native-app-setup-script` + `make -C native_app build`):

- 39 maps present, `sourcesContent` included, paths re-rooted, **no** absolute or `..` leaks
- 139 `sourceMappingURL` comments in `modules.sql`, all resolved to `sourcemaps/`
- the three tiler UDTFs that inline a placeholder **mid-line** (`_GEOJSON_TO_MVT_TILE`, `_GEOJSON_TO_BINARY_TILE`, `_GEOJSON_TO_BINARY_SPATIAL_INDEX_TILE`) still parse — rollup emits a newline after the comment, so the following `},` lands on its own line
- both `PUT` statements resolve correctly (`make -n`)
- library test suites pass: core 13/13, premium 16/16 (179 tests) — the `UNIT_TEST` umd build is unaffected
- `make lint` clean in core

## Two things for the reviewer

**1. The first `PUT` glob changed from `/*` to `/*.*`.** `PUT` cannot upload directories and does not recurse, so the original glob would now also match `sourcemaps/`. `*.*` matches files only — all six package files have extensions — and the sub-directory gets its own statement. Flagging it because it is the one behaviour that cannot be verified without a real stage, and because an extension-less package file added later would be silently skipped. It now carries a comment saying so, because tidying it to `*` would break the upload. Happy to switch to explicit per-file statements if you prefer.

**2. Package size.** The premium package goes to ~11 MB (4.5 MB `modules.sql` + ~7 MB of maps). No documented size limit for a version's files was found, but it is a real jump. If it becomes a problem, `sourcemapExcludeSources` cuts the maps by ~70% — at the cost of no longer satisfying "recover the un-minified code" from the map alone.

## Rollout

- This does not take effect for the premium package until AT bumps the submodule — **deliberately not included here**, per the convention of keeping the bump separate.
- **Already validated in production.** Patch 28 of the premium package was built from this branch and submitted, and Snowflake's scan **approved it automatically** — no appeal spent. That answers the open question this PR was originally written against: the scan does resolve `//# sourceMappingURL=` out of JavaScript embedded in a `.sql` file, and it follows the relative `sourcemaps/` sub-path. Snowflake also confirmed in writing that files in the version root which the manifest does not declare travel with the app but are not readable by consumers.
- The layout stays one variable (`APP_SOURCE_MAPS_DIR`) and one `sed`, if Snowflake ever wants a different arrangement.

Story: sc-571986

### DIFF
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -6,7 +6,6 @@ DIST_DIR ?= $(ROOT_DIR)/dist
 BUILD_DIR ?= $(ROOT_DIR)/build
 ESLINTRC_DIR ?= $(ROOT_DIR)/../..
 COMMON_DIR = $(ROOT_DIR)/common
-LIBS_BUILD_DIR ?= libraries/javascript/build
 PACKAGE_VERSION ?= $(shell cat $(ROOT_DIR)/version)
 PACKAGE_NAME ?= carto-analytics-toolbox-core-snowflake-$(PACKAGE_VERSION)
 
@@ -93,7 +92,7 @@ build-native-app-setup-script:
 # Point each bundle's sourceMappingURL at the package sub-directory, before they are inlined
 prefix-library-sourcemap-urls:
 	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
-	set -e; for f in $(LIBS_BUILD_DIR)/*.js; do \
+	set -e; for f in libraries/javascript/build/*.js; do \
 		sed -i.bak 's|//# sourceMappingURL=|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f; \
 		rm -f $$f.bak; \
 	done
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -6,6 +6,7 @@ DIST_DIR ?= $(ROOT_DIR)/dist
 BUILD_DIR ?= $(ROOT_DIR)/build
 ESLINTRC_DIR ?= $(ROOT_DIR)/../..
 COMMON_DIR = $(ROOT_DIR)/common
+LIBS_BUILD_DIR ?= libraries/javascript/build
 PACKAGE_VERSION ?= $(shell cat $(ROOT_DIR)/version)
 PACKAGE_NAME ?= carto-analytics-toolbox-core-snowflake-$(PACKAGE_VERSION)
 
@@ -92,7 +93,7 @@ build-native-app-setup-script:
 # Point each bundle's sourceMappingURL at the package sub-directory, before they are inlined
 prefix-library-sourcemap-urls:
 	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
-	for f in libraries/javascript/build/*.js; do \
+	set -e; for f in $(LIBS_BUILD_DIR)/*.js; do \
 		sed -i.bak 's|//# sourceMappingURL=|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f; \
 		rm -f $$f.bak; \
 	done
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -79,7 +79,7 @@ build-modules:
 
 build-native-app-setup-script:
 	rm -rf $(BUILD_DIR)
-	$(MAKE) build-libraries
+	$(MAKE) build-libraries sourcemap=1
 	$(MAKE) prefix-library-sourcemap-urls
 	mkdir -p $(BUILD_DIR)
 	$(MAKE) -C modules build SF_SCHEMA=@@SF_SCHEMA@@
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -84,10 +84,10 @@ build-native-app-setup-script:
 	mkdir -p $(BUILD_DIR)
 	$(MAKE) -C modules build SF_SCHEMA=@@SF_SCHEMA@@
 	cp modules/build/modules.sql $(BUILD_DIR)
-	$(MAKE) -C modules build-native-app-setup-script
-	cp modules/build/setup_script.sql $(BUILD_DIR)
 	$(MAKE) -C modules build-source-review
 	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
+	$(MAKE) -C modules build-native-app-setup-script
+	cp modules/build/setup_script.sql $(BUILD_DIR)
 
 # The app package keeps the source maps in a sub-directory, so the
 # sourceMappingURL comment rollup wrote next to each bundle needs the same
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -89,7 +89,7 @@ build-native-app-setup-script:
 	$(MAKE) -C modules build-source-review
 	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
 
-# Point each bundle's sourceMappingURL at the package sub-directory, before they are inlined
+# Must run before the bundles are inlined into modules.sql
 prefix-library-sourcemap-urls:
 	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
 	set -e; for f in libraries/javascript/build/*.js; do \
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -93,7 +93,7 @@ build-native-app-setup-script:
 prefix-library-sourcemap-urls:
 	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
 	set -e; for f in libraries/javascript/build/*.js; do \
-		sed -i.bak 's|//# sourceMappingURL=|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f; \
+		sed -i.bak 's|//# sourceMappingURL=\($(APP_SOURCE_MAPS_DIR)/\)*|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f; \
 		rm -f $$f.bak; \
 	done
 
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -13,7 +13,7 @@ include $(COMMON_DIR)/Makefile
 
 .SILENT:
 
-.PHONY: help lint lint-fix build deploy test benchmark remove clean create-package deploy-native-app-package deploy-native-app
+.PHONY: help lint lint-fix build deploy test benchmark remove clean create-package prefix-library-sourcemap-urls deploy-native-app-package deploy-native-app
 
 help:
 	echo "Available targets: lint build deploy test remove clean create-package deploy-native-app-package deploy-native-app"
@@ -80,11 +80,24 @@ build-modules:
 build-native-app-setup-script:
 	rm -rf $(BUILD_DIR)
 	$(MAKE) build-libraries
+	$(MAKE) prefix-library-sourcemap-urls
 	mkdir -p $(BUILD_DIR)
 	$(MAKE) -C modules build SF_SCHEMA=@@SF_SCHEMA@@
 	cp modules/build/modules.sql $(BUILD_DIR)
 	$(MAKE) -C modules build-native-app-setup-script
 	cp modules/build/setup_script.sql $(BUILD_DIR)
+	$(MAKE) -C modules build-source-review
+	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
+
+# The app package keeps the source maps in a sub-directory, so the
+# sourceMappingURL comment rollup wrote next to each bundle needs the same
+# prefix before the bundles are inlined into modules.sql.
+prefix-library-sourcemap-urls:
+	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
+	for f in libraries/javascript/build/*.js; do \
+		sed -i.bak 's|//# sourceMappingURL=|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f; \
+		rm -f $$f.bak; \
+	done
 
 deploy:
 	$(MAKE) build-libraries
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -85,10 +85,10 @@ build-native-app-setup-script:
 	mkdir -p $(BUILD_DIR)
 	$(MAKE) -C modules build SF_SCHEMA=@@SF_SCHEMA@@
 	cp modules/build/modules.sql $(BUILD_DIR)
-	$(MAKE) -C modules build-source-review
-	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
 	$(MAKE) -C modules build-native-app-setup-script
 	cp modules/build/setup_script.sql $(BUILD_DIR)
+	$(MAKE) -C modules build-source-review
+	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
 
 # Point each bundle's sourceMappingURL at the package sub-directory, before they are inlined
 prefix-library-sourcemap-urls:
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -89,7 +89,6 @@ build-native-app-setup-script:
 	$(MAKE) -C modules build-source-review
 	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
 
-# Must run before the bundles are inlined into modules.sql
 prefix-library-sourcemap-urls:
 	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
 	set -e; for f in libraries/javascript/build/*.js; do \
--- a/clouds/snowflake/Makefile
+++ b/clouds/snowflake/Makefile
@@ -89,9 +89,7 @@ build-native-app-setup-script:
 	$(MAKE) -C modules build-native-app-setup-script
 	cp modules/build/setup_script.sql $(BUILD_DIR)
 
-# The app package keeps the source maps in a sub-directory, so the
-# sourceMappingURL comment rollup wrote next to each bundle needs the same
-# prefix before the bundles are inlined into modules.sql.
+# Point each bundle's sourceMappingURL at the package sub-directory, before they are inlined
 prefix-library-sourcemap-urls:
 	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
 	for f in libraries/javascript/build/*.js; do \
--- a/clouds/snowflake/common/build_source_review.js
+++ b/clouds/snowflake/common/build_source_review.js
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 
-// Build the source review index for the native app package
-// and check every inlined library ships a usable source map
+// Build the source review index and check every inlined library ships a usable source map
 
 // ./build_source_review.js modules --output=build --libs_build_dir=../libraries/javascript/build --source_maps_dir=sourcemaps
 
@@ -31,8 +30,7 @@ for (let inputDir of inputDirs) {
             files.forEach(file => {
                 if (file.endsWith('.sql')) {
                     const name = path.parse(file).name;
-                    // Strip SQL comments as build_modules.js and list_libraries.js do, so a
-                    // commented-out placeholder is not recorded as an inlined library
+                    // Strip comments as build_modules.js does: a commented-out placeholder is not inlined
                     const content = fs.readFileSync(path.join(moduledir, file)).toString().replace(/--.*\n/g, '');
                     const libraries = [... new Set(content.match(/@@SF_LIBRARY_[A-Z0-9_]+@@/g) || [])]
                         .map(l => l.replace('@@SF_LIBRARY_', '').replace('@@', '').toLowerCase());
@@ -45,7 +43,6 @@ for (let inputDir of inputDirs) {
     });
 }
 
-// Invert into library -> functions
 const libraries = {};
 functions.forEach(f => {
     f.libraries.forEach(library => {
@@ -75,13 +72,10 @@ libraryNames.forEach(library => {
     try {
         map = JSON.parse(fs.readFileSync(mapPath).toString());
     } catch (e) {
-        // Report it rather than dying with a raw parse error, so the build failure names the file
         errors.push(`source map "${library}.js.map" is not valid JSON: ${e.message}`);
         return;
     }
-    // Every original must be recoverable from the map alone, which is what the index promises
-    // the reviewer. A non-empty sourcesContent is not enough: a map with no sources at all, or
-    // with null entries, recovers nothing while still satisfying a presence check.
+    // The index promises every original is recoverable from the map alone, so presence is not enough
     const sources = map.sources || [];
     const contents = map.sourcesContent || [];
     const missing = sources.filter((_, i) => typeof contents[i] !== 'string' || contents[i] === '');
@@ -103,7 +97,6 @@ if (errors.length) {
     process.exit(1);
 }
 
-// The document, derived entirely from this build and aimed at whoever reviews the package
 const functionRows = functions
     .sort((a, b) => a.name.localeCompare(b.name))
     .map(f => `| \`${f.name}\` | ${f.libraries.sort().map(l => `\`${sourceMapsDir}/${l}.js.map\``).join(', ')} |`)
--- a/clouds/snowflake/common/build_source_review.js
+++ b/clouds/snowflake/common/build_source_review.js
@@ -13,11 +13,8 @@
 
 const fs = require('fs');
 const path = require('path');
-const crypto = require('crypto');
 const { execSync } = require('child_process');
 const argv = require('minimist')(process.argv.slice(2));
-
-const sha256 = contents => crypto.createHash('sha256').update(contents).digest('hex');
 
 const inputDirs = argv._[0] && argv._[0].split(',');
 const outputDir = argv.output || 'build';
@@ -85,11 +82,7 @@ const entries = Object.keys(libraries).sort().map(library => {
     }
     return {
         library,
-        functions: [... new Set(libraries[library])].sort(),
-        bundleSize: bundle.length,
-        mapSize: fs.statSync(mapPath).size,
-        mapSha256: sha256(fs.readFileSync(mapPath)),
-        sources: (map.sources || []).length
+        functions: [... new Set(libraries[library])].sort()
     };
 });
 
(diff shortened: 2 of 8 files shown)


vdelacruzb:
- chore sf ship JS source maps in the Snowflake [sc-571986]

Jesus89 (commented):
Ten notes inline, roughly by severity. The first one matters most: as written, `check-package-sourcemaps` exits 0 when `modules.sql` references no map at all, so the guard cannot catch the case it was added for.

I have local patches for all ten if that's useful.

clouds/snowflake/native_app/Makefile
```diff
@@ -39,11 +40,24 @@ build:
...
+	cp $(BUILD_DIR)/SOURCE_REVIEW.md $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/
+	mkdir -p $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/$(APP_SOURCE_MAPS_DIR)
+	cp $(LIBS_BUILD_DIR)/*.js.map $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/$(APP_SOURCE_MAPS_DIR)/
+	$(MAKE) check-package-sourcemaps
 
 
+# A wrong prefix or a half-applied rename still builds, so check the references resolve
+check-package-sourcemaps:
+	set -e; pkg=$(DIST_DIR)/$(APP_PACKAGE_DIST_NAME); \
+	for map in `grep -o '//# sourceMappingURL=[^ ]*' $$pkg/modules.sql | sed 's|.*=||' | sort -u`; do \
```
Jesus89: The `grep` runs in a command substitution feeding a `for` list, so when it matches nothing the list is empty, the loop body never runs, and `set -e` does not fire. The check therefore passes when `modules.sql` has *zero* `sourceMappingURL` references — the regression it exists to catch. Verified: a `modules.sql` containing only `SELECT 1;` prints `CHECK PASSED` and exits 0.

Capture first and require at least one: `maps=\`grep ...\`; test -n "$$maps" || { echo "ERROR: ..."; exit 1; }`.

clouds/snowflake/common/rollup.config.js
```diff
@@ -30,6 +30,18 @@ export default {
     input,
     output: {
         file: process.env.OUTPUT,
+        sourcemap: Boolean(process.env.SOURCEMAP),
```
Jesus89: `Boolean('0') === true`, so `make build-libraries sourcemap=0` (or `sourcemap=false`) still emits `.map` files and the `sourceMappingURL` comment — which then gets inlined into `modules.sql` on the plain `deploy` / `create-package` paths, where no maps ship. Use `process.env.SOURCEMAP === '1'`.

clouds/snowflake/common/build_source_review.js
```diff
@@ -0,0 +1,119 @@
...
+    }
+    if (!fs.readFileSync(bundlePath).toString().includes('//# sourceMappingURL=')) {
+        errors.push(`bundle "${library}.js" has no sourceMappingURL comment: enable "sourcemap" in the rollup config`);
+    }
+    if (!fs.existsSync(mapPath)) {
+        errors.push(`library "${library}" has no source map at ${mapPath}`);
+        return;
+    }
+    const map = JSON.parse(fs.readFileSync(mapPath).toString());
+    if (!map.sourcesContent || !map.sourcesContent.length) {
```
Jesus89: This only asserts the array is non-empty. A map with 58 `sources` and `sourcesContent: ["...", null, null, ...]` passes while 57 originals are unrecoverable, and SOURCE_REVIEW.md still tells the reviewer the source can be recovered from the map alone. Compare lengths and require every entry to be a non-empty string.

clouds/snowflake/Makefile
```diff
@@ -79,12 +79,23 @@ build-modules:
...
 	$(MAKE) -C modules build-native-app-setup-script
 	cp modules/build/setup_script.sql $(BUILD_DIR)
+	$(MAKE) -C modules build-source-review
+	cp modules/build/SOURCE_REVIEW.md $(BUILD_DIR)
+
+# Point each bundle's sourceMappingURL at the package sub-directory, before they are inlined
+prefix-library-sourcemap-urls:
+	echo "Prefixing library sourceMappingURL comments with $(APP_SOURCE_MAPS_DIR)/..."
+	set -e; for f in libraries/javascript/build/*.js; do \
+		sed -i.bak 's|//# sourceMappingURL=|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f; \
```
Jesus89: Not idempotent: the pattern also matches an already-prefixed comment, so a second run produces `sourcemaps/sourcemaps/h3.js.map` (verified). The target is `.PHONY` and edits a shared artifact in place, so `make prefix-library-sourcemap-urls` twice, then `deploy` / `create-package`, ships the broken URL with no guard on that path. Anchoring fixes it: `s|//# sourceMappingURL=\(sourcemaps/\)*|...|`.

clouds/snowflake/common/build_source_review.js
```diff
@@ -0,0 +1,119 @@
...
+    const sqldir = path.join(inputDir, 'sql');
+    const modules = fs.readdirSync(sqldir);
+    modules.forEach(module => {
+        const moduledir = path.join(sqldir, module);
+        if (fs.statSync(moduledir).isDirectory()) {
+            const files = fs.readdirSync(moduledir);
+            files.forEach(file => {
+                if (file.endsWith('.sql')) {
+                    const name = path.parse(file).name;
+                    const content = fs.readFileSync(path.join(moduledir, file)).toString();
```
Jesus89: `build_modules.js:73` and `list_libraries.js:57` both `.replace(/--.*\n/g, '')` before matching placeholders; this reads the raw file. A commented-out `-- @@SF_LIBRARY_FOO@@` then makes this script record `foo` as inlined while the builders never build it — the native-app build fails with `library "foo" ... does not exist`. Same stripping here keeps the three in agreement.

clouds/snowflake/modules/Makefile
```diff
@@ -86,6 +86,12 @@ build-share: $(NODE_MODULES_DEV)
 		--output=$(BUILD_DIR) --libs_build_dir=$(LIBS_BUILD_DIR) --diff="$(diff)" \
 		--modules=$(modules) --functions=$(functions) --production=$(production) --nodeps=$(nodeps) --dropfirst=$(dropfirst)
 
+build-source-review: $(NODE_MODULES_DEV)
```
Jesus89, line 89: Unlike `build`, `build-share` and `build-native-app-setup-script`, this target never creates `$(BUILD_DIR)`. Run on its own on a clean checkout it validates fine and then dies with a raw `ENOENT ... build/SOURCE_REVIEW.md` from `writeFileSync`, bypassing the script's own `ERROR:` path. It only works today because the chain runs `build-native-app-setup-script` first. `mkdir -p $(BUILD_DIR)` covers it.

clouds/snowflake/common/build_source_review.js
```diff
@@ -0,0 +1,119 @@
...
+// ./build_source_review.js modules --output=build --libs_build_dir=../libraries/javascript/build
+
+const fs = require('fs');
+const path = require('path');
+const argv = require('minimist')(process.argv.slice(2));
+
+const inputDirs = argv._[0] && argv._[0].split(',');
+const outputDir = argv.output || 'build';
+const libsBuildDir = argv.libs_build_dir || '../libraries/javascript/build';
+const sourceMapsDir = argv.source_maps_dir || 'source_review';
```
Jesus89: Two defaults for one directory: `APP_SOURCE_MAPS_DIR ?= sourcemaps` in common/Makefile:6 vs `source_review` here. An empty `--source_maps_dir=` also falls through to it, and the usage line above omits the flag entirely — either way the index tells the reviewer the maps are in `source_review/` while the package ships them in `sourcemaps/`. `check-package-sourcemaps` validates the bundle comments, not the index, so it ships silently.

clouds/snowflake/common/build_source_review.js
```diff
@@ -0,0 +1,119 @@
...
+    process.exit(1);
+}
+
+// The document, derived entirely from this build and aimed at whoever reviews the package
+const functionRows = functions
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .map(f => `| \`${f.name}\` | ${f.libraries.sort().map(l => `\`${sourceMapsDir}/${l}.js.map\``).join(', ')} |`)
+    .join('\n');
+
+const document = `# Source review
```
Jesus89, line 105: The generated document is missing most of what the PR description advertises: per-library map bytes / source count, bundle bytes + sha256, `inlined into N function(s)`, the `library → functions` direction, and build provenance (commit, node, rollup, command). The `libraries` inversion built at lines 42-48 is currently only used inside error strings.

clouds/snowflake/native_app/Makefile
```diff
@@ -39,11 +40,24 @@ build:
 	sed 's/@@VERSION@@/$(APP_VERSION)/g' $(APP_DIR)/manifest.yml > $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/manifest.yml
 	cp $(APP_DIR)/README.md $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/
 	cp $(APP_DIR)/get_modules_sql_from_stage.py $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/
+	cp $(BUILD_DIR)/SOURCE_REVIEW.md $(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/
```
Jesus89, line 43: `manifest.yml` declares only `readme: README.md`, and `native_app/README.md` never mentions source maps or this file — so the surface Snowflake renders to a reviewer has no pointer to the evidence. A couple of lines in the README (both repos) is what makes it discoverable.

clouds/snowflake/native_app/Makefile
```diff
@@ -39,11 +40,24 @@ build:
...
+	for map in `grep -o '//# sourceMappingURL=[^ ]*' $$pkg/modules.sql | sed 's|.*=||' | sort -u`; do \
+		test -f $$pkg/$$map || { echo "ERROR: modules.sql references $$map, which is not in the package"; exit 1; }; \
+	done
+
 deploy-app-package:
 	$(COMMON_DIR)/run-query.js "CREATE STAGE IF NOT EXISTS $(APP_STAGE_NAME);"
-	$(COMMON_DIR)/run-query.js "PUT file://$(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/* @$(APP_STAGE_LOCATION) overwrite=true auto_compress=false;"
+# PUT cannot upload directories, so this glob takes files only and the next statement the maps
+	$(COMMON_DIR)/run-query.js "PUT file://$(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/*.* @$(APP_STAGE_LOCATION) overwrite=true auto_compress=false;"
+	$(COMMON_DIR)/run-query.js "PUT file://$(DIST_DIR)/$(APP_PACKAGE_DIST_NAME)/$(APP_SOURCE_MAPS_DIR)/* @$(APP_STAGE_LOCATION)/$(APP_SOURCE_MAPS_DIR) overwrite=true auto_compress=false;"
```
Jesus89, line 67: `overwrite=true` replaces same-named files but never deletes, and for non-production builds `APP_STAGE_LOCATION` carries no version prefix, so the prefix is shared across every build. Rename or drop a library and its old `.js.map` stays on the stage; `ALTER APPLICATION PACKAGE ... USING @$(APP_STAGE_LOCATION)` then snapshots an orphan map into the next patch — unreferenced and absent from SOURCE_REVIEW.md. A `REMOVE @$(APP_STAGE_LOCATION)/$(APP_SOURCE_MAPS_DIR);` before the PUTs handles it.

Jesus89 (changes requested):

vdelacruzb:
Thanks — all ten were real. Nine are fixed in d2356f3e; the tenth is a stale PR description rather than a code problem, and I have updated it.

I reproduced each failure before fixing it and re-tested after, since several were the kind that pass silently.

| # | Finding | Fix | Verified |
|---|---|---|---|
| 1 | Guard exits 0 with zero map references | Capture the list, require non-empty | `modules.sql` of `SELECT 1;` now exits **2**; real package still exits 0 |
| 2 | `Boolean('0') === true` | `process.env.SOURCEMAP === '1'` | `sourcemap=0` now writes **0** maps and no comment |
| 3 | `sourcesContent` only checked non-empty | Compare lengths, require every entry to be a non-empty string | Injected one `null` into a 2-source map → fails naming the source |
| 4 | Prefix `sed` not idempotent | Anchor with `\(sourcemaps/\)*` | Target run twice → single prefix |
| 5 | Raw SQL read, no `--` stripping | Same `.replace(/--.*\n/g, '')` as the other two builders | Commented-out placeholder no longer recorded |
| 6 | `build-source-review` never created `BUILD_DIR` | `mkdir -p` | Standalone on a wiped `modules/build` now succeeds |
| 7 | Two defaults for one directory | Script defaults to `sourcemaps`, with a note to keep it in sync | Index now emits `sourcemaps/statistics.js.map` |
| 9 | README had no pointer to the evidence | Section in `native_app/README.md` | — |
| 10 | Orphan maps snapshotted from a shared stage prefix | `REMOVE @$(APP_STAGE_LOCATION)/$(APP_SOURCE_MAPS_DIR);` before the PUTs | `make -n` shows REMOVE ordered before both PUTs |

Your first one was the important one, and it was worse than it looked: the guard was added precisely to catch a wrong or missing prefix, and a wrong prefix produces zero matches — so it was blind to its own reason for existing. Same for #3, where the index tells the reviewer the source is recoverable from the map alone, which a `null` entry silently makes false.

**On #8 I would rather not re-add the content.** The sizes, per-bundle SHA-256s and build provenance were in an earlier revision and were removed deliberately, after review of the customer-facing version of this document. Two reasons: a hash of a map printed inside the same package as that map is self-referential, and the reproducibility it was supporting does not work for a reviewer who cannot rebuild a private repo. What is left is the review path — how to read a map, and which map covers which function. The `library → functions` direction went the same way, since the function table answers the question in the direction a reviewer asks it.

So the PR description was the thing out of date, not the generator. I have rewritten that section, corrected the two `source_review/` references, and updated the guard's description to match what it now checks.

One note on your side-observation that the `libraries` inversion is only used in error strings: it is also what builds `libraryNames`, which drives the per-library validation loop and the summary count. I have left it.

Full lint clean in core, premium library tests 16/16 (179 tests), and both packages rebuilt and re-verified end to end.

Happy to take your local patches if any of these differ from what you had in mind — particularly #4, where you may have anchored it differently.

clouds/snowflake/native_app/Makefile
vdelacruzb: Pushing back on this one, with evidence from the shipped package.

Your premise is right — `manifest.yml` declares only `readme: README.md`. But that cuts the other
way: a declared readme artifact is the surface Snowflake renders **to consumers**, not just to a
reviewer. Adding minification and source-map mechanics there puts build internals into an end-user
document, and invites a "why is your code minified?" question from customers who would otherwise
never ask.

And the pointer turns out not to be needed for discoverability. The patch-28 build we already
shipped and got **approved automatically** contains:

- `README.md` with zero mention of source maps
- `SOURCE_REVIEW.md` at the package root
- 39 maps under `sourcemaps/`

So the scanner resolved `//# sourceMappingURL=` out of the SQL and found the maps on its own, with no
README pointer whatsoever. That is the case your finding predicted would fail, and it passed.

`SOURCE_REVIEW.md` already carries everything the README section said and more — the minification, the
maps, `sourcesContent`, why one map covers several functions, how to read a map, and the
function → map table. It sits at the package root and is undeclared in the manifest, so per
Snowflake's written confirmation to us it travels with the app without being consumer-readable.
That is the right home for it.

So I have reverted the README section (58701664). Happy to put a single neutral sentence back if you
want cover for a *manual* appeal review rather than the automated scan — but I would rather not
carry it purely on spec, given the consumer-facing cost.

I had also opened CartoDB/analytics-toolbox#1192 to mirror this into the premium README, since
`APP_DIR = $(MAKE_DIR)` there means the premium package ships its own copy. Closed for the same
reason.

clouds/snowflake/common/build_source_review.js
vdelacruzb: You were more right than my first fix. I only made the two defaults *agree*, which fixed the symptom
(the index no longer names a directory the package does not ship) but left the duplication itself —
an absent or empty `--source_maps_dir` still silently picked a value the script chose for itself, and
the usage line still omitted the flag.

Now there is no second default (ee33403d):

```
const sourceMapsDir = argv.source_maps_dir;  // no default: APP_SOURCE_MAPS_DIR is the only source of truth

if (!sourceMapsDir) {
    console.log('ERROR: --source_maps_dir is required and must match APP_SOURCE_MAPS_DIR');
    process.exit(1);
}
```

Verified: flag absent and `--source_maps_dir=` both now exit 1 with that message; with the flag the
normal path is unchanged (25 libraries, 51 functions, index pointing at 25 maps under `sourcemaps/`).
Usage line updated to name the flag. eslint clean.

clouds/snowflake/native_app/Makefile
vdelacruzb: Fixed in d2356f3e. Captured first, then required non-empty, as you suggested:

```make
maps=`grep -o '//# sourceMappingURL=[^ ]*' $$pkg/modules.sql | sed 's|.*=||' | sort -u`; \
test -n "$$maps" || { echo "ERROR: modules.sql references no source map at all"; exit 1; }; \
```

Verified: a package whose `modules.sql` is just `SELECT 1;` now prints
`ERROR: modules.sql references no source map at all` and exits 1 (make reports 2). The real package
still passes.

This was the important one, and worse than it first looks: the guard exists to catch a wrong or
missing prefix, and a wrong prefix produces *zero* matches — so it was blind to precisely the
regression it was added for.

clouds/snowflake/common/rollup.config.js
vdelacruzb: Fixed in d2356f3e — `rollup.config.js:33` is now `sourcemap: process.env.SOURCEMAP === '1'`.

Confirmed the premise: `Boolean("0") === true`.

So `make build-libraries sourcemap=0` now writes zero maps and emits no `sourceMappingURL` comment,
which closes the path you identified — a comment inlined into `modules.sql` on the plain `deploy` /
`create-package` runs, where no maps ship alongside it.

clouds/snowflake/common/build_source_review.js
vdelacruzb: Fixed in d2356f3e, and applying your bug class to the rest of the function turned up two more, both
in 1ac31a38. The check now compares lengths, requires every entry to be a non-empty string, requires
`sources` itself to be non-empty, and wraps the `JSON.parse`.

Verified against the real script, corrupting `accessors.js.map` three ways:

```
one null entry   ERROR: source map "accessors.js.map" is missing the content of 1 of 6 sources,
                 starting with ".../@turf/helpers/dist/esm/index.js"
sources: []      ERROR: source map "accessors.js.map" lists no sources, so no original can be
                 recovered from it
not JSON         ERROR: source map "accessors.js.map" is not valid JSON: Unexpected token 'o'
intact           Write .../SOURCE_REVIEW.md (25 libraries, 51 functions)
```

The `sources: []` case is the one worth flagging: two empty arrays satisfy *both* the length compare
and the per-entry test, so a map from which nothing can be recovered would have passed every check
while `SOURCE_REVIEW.md` told the reviewer the source is recoverable from the map alone. Same shape as
this finding, one level down. The corrupt-JSON case was the same shape as your `mkdir` one — it died
with a raw `SyntaxError` instead of the script's own `ERROR:` path, so the failure never named the file.

clouds/snowflake/Makefile
vdelacruzb: Fixed in d2356f3e with your anchoring:

```make
sed -i.bak 's|//# sourceMappingURL=\($(APP_SOURCE_MAPS_DIR)/\)*|//# sourceMappingURL=$(APP_SOURCE_MAPS_DIR)/|' $$f
```

Verified: two consecutive runs leave `//# sourceMappingURL=sourcemaps/h3.js.map` — previously the
second produced `sourcemaps/sourcemaps/...`.

clouds/snowflake/common/build_source_review.js
vdelacruzb: Fixed in d2356f3e — the same `.replace(/--.*\n/g, '')` the other two builders use, so all three now
read the SQL identically.

Verified: appending `-- @@SF_LIBRARY_NOSUCHLIB@@` to a module SQL file no longer records the library,
and the run completes clean. Without the stripping that produced
`library "nosuchlib" is inlined by ... but ... does not exist` and failed the native-app build, exactly
as you described.

clouds/snowflake/modules/Makefile
vdelacruzb: Fixed in d2356f3e — `mkdir -p $(BUILD_DIR)` ahead of the script.

Verified standalone on a wiped build dir:

```
rm -rf modules/build && make -C modules build-source-review
  Building source review...
  Write .../modules/build/SOURCE_REVIEW.md (25 libraries, 51 functions)
```

Previously that died with a raw `ENOENT ... build/SOURCE_REVIEW.md` out of `writeFileSync`, bypassing
the script's own `ERROR:` path — same class as the corrupt-map case, where an unhandled throw hides
which file was at fault.

clouds/snowflake/common/build_source_review.js
vdelacruzb: This one I would rather not act on, because the PR description was the stale artifact, not the
generator — and I have rewritten it.

The per-library sizes, bundle SHA-256s and build provenance were in an earlier revision and were
removed deliberately, after review of the customer-facing version of this document. Two reasons: a
hash of a file printed inside the same package as that file is self-referential and proves nothing to
a reader who already has both; and the reproducibility it was there to support does not work for a
reviewer who cannot rebuild a private repo. What is left is the review path — how to read a map, and
which map covers which function.

The `library → functions` direction went the same way: the function table already answers that
question in the direction a reviewer asks it ("I am looking at this function, where is its source"),
and carrying both directions in one generated document means two things to keep consistent.

On the side observation that the `libraries` inversion is only used in error strings — it also builds
`libraryNames`, which drives the per-library validation loop and the summary count, so it is load
bearing. Left as is.

Happy to be overruled on the sizes if you think a reviewer benefits, but I did not want to re-add
content that a previous round of review took out.

clouds/snowflake/native_app/Makefile
vdelacruzb: Fixed in d2356f3e — `REMOVE @$(APP_STAGE_LOCATION)/$(APP_SOURCE_MAPS_DIR);` at `native_app/Makefile:62`,
ordered before both PUTs (confirmed with `make -n`).

Worth recording that your finding already applies to what we shipped: the dev stage prefix was shared
across several iterations of this branch, including an earlier `source_review/` naming, so patch 28
may well carry orphan maps that are unreferenced and absent from `SOURCE_REVIEW.md`. We decided to
leave that release alone — the orphans are source maps rather than minified JavaScript, so they cannot
affect the security scan, and the patch is already approved. Future patches are covered by the REMOVE.

Jesus89:
All ten threads resolved. Eight fixes verified in the diff, and I'm taking both pushbacks — the README one is settled by the patch-28 evidence (the scanner resolved the maps out of the SQL with no pointer), and I'd rather not re-add document content a previous round deliberately removed.

One residual item, a follow-up to the `REMOVE` fix rather than a change request on it.

`REMOVE @$(APP_STAGE_LOCATION)/$(APP_SOURCE_MAPS_DIR);` clears `sourcemaps/`, which covers every future build. It does not clear the earlier `source_review/` prefix this branch used before `237016d0` renamed the directory. Those maps are still sitting on the shared dev stage, and since `ALTER APPLICATION PACKAGE ... USING @$(APP_STAGE_LOCATION)` snapshots the whole stage, they get picked up by every subsequent dev patch — not only patch 28. Agreed that leaving patch 28 alone is fine (they're maps, not minified JS, so the scan is unaffected), but the carry-forward is worth stopping now rather than shipping orphans indefinitely.

One manual statement against the dev stage does it:

```sql
REMOVE @<APP_STAGE_LOCATION>/source_review;
```

Worth confirming with `LIST @<APP_STAGE_LOCATION>;` afterwards that nothing else from the earlier iterations is left up there.

Separately, and lower priority: `check-package-sourcemaps` validates `modules.sql` → map, but not the reverse. A map present in the package that nothing references still passes. The stage side is covered by the `REMOVE` now, so this is only about the packaging step, and I don't think it needs to block here.

Approving once the stage cleanup is done or you tell me you'd rather track it separately.

vdelacruzb (commented):
Thanks — and the stage item is done, with a result that lets you approve without any cleanup: **there is nothing up there to remove.**

I checked every installer stage — the production one, the shared dev one, and two older dev stages — with `LIST @<stage> PATTERN='.*source_review.*'`. **Zero matches on all four.**

Patch 28's snapshot is exactly what it should be:

```
v1_26_2/             README.md  SOURCE_REVIEW.md  get_modules_sql_from_stage.py
                     manifest.yml  modules.sql  setup_script.sql      (6 files)
v1_26_2/sourcemaps/  39 maps
```

Your reasoning was right — `REMOVE` does clear only `sourcemaps/`, and the dev prefix is unversioned and shared. The premise just did not hold: `237016d0` renamed the directory **before** the first deploy, so every `PUT` ever made used `sourcemaps/`. The `source_review/` prefix only existed in local builds on my machine.

One incidental find: the production stage has a handful of files at its root, outside any version prefix, left from pre-versioning deploys. Harmless, since `ALTER APPLICATION PACKAGE … USING @<stage>/<version>` points at the version prefix, so they are never snapshotted.

## I took the packaging one too, rather than defer it

You were right that it needs doing, and the mechanism is worse than "a stale file": `cp $(LIBS_BUILD_DIR)/*.js.map` copies a map for **every built library**, referenced or not. So a library still present in `libraries/javascript/libs/` but no longer inlined by any SQL ships an unreferenced map, while `SOURCE_REVIEW.md` says nothing about it. Nothing to do with staleness — the libs build dir is cleaned on every build.

The guard now checks both directions (98e475fe), verified by triggering it:

```
clean package      exit 0
stray orphan map   ERROR: sourcemaps/orphan_lib.js.map ships in the package
                   but nothing references it            exit 2
restored           exit 0
```

Snowflake CI is running on that commit; nothing else changed.

Jesus89 (approved):

vdelacruzb:
Snowflake CI green on `98e475fe`.

